PR #53, opened 2026-07-31
Add @kavo/prisma adapter package alongside @kavo/typeorm

### LINKED ISSUE #52: Add @kavo/prisma adapter package alongside @kavo/typeorm
## Context

Kavo currently ships one ORM adapter, `@kavo/typeorm`, which implements core's `RepositoryAdapter` and `CrudInfrastructure` seams (ADR-0011) so core stays ORM-independent (ADR-0001, ADR-0005). Prisma is the other ORM Kavo should support, wired in the same way: a new adapter package that translates core's ORM-independent contracts to Prisma Client, with zero changes to core and zero coupling between adapters.

## Acceptance criteria

- [ ] New package `packages/orms/prisma`, workspace name `@kavo/prisma`, picked up automatically by the existing `packages/orms/*` glob in `pnpm-workspace.yaml` (no workspace config change needed).
- [ ] `metadata.ts`: `buildEntityMetadata(prismaClient, entity)` → core's `EntityMetadata` (fields, relations, `idField`, `softDeleteField`), derived from Prisma's runtime DMMF (`Prisma.dmmf.datamodel.models`). Scalar → `FieldKind` mapping for String/Int/Float/Boolean/DateTime/Json/Enum/BigInt/Decimal. Composite primary keys rejected with `ConfigurationException`, matching `@kavo/typeorm`'s single-primary-column constraint.
- [ ] `filter-translator.ts`: Filter AST → Prisma `where` object — AND/OR/NOT groups, relation-path conditions via nested `{ relation: { field: ... } }`, and all of `EQ/NE/GT/GTE/LT/LTE/IN/NOT_IN/LIKE/ILIKE/BETWEEN/IS_NULL/IS_NOT_NULL` mapped to Prisma's filter operators (`ILIKE` via `mode: "insensitive"`, `BETWEEN` via combined `gte`+`lte`). `assertNever` over the operator union, matching the TypeORM translator's exhaustiveness check.
- [ ] `error-mapping.ts`: `PrismaClientKnownRequestError` codes → Kavo exceptions — P2002 (unique) and P2003/P2014 (FK) → `ConflictException`, P2034 (transaction conflict) → `TransactionException` (retryable), everything else → `PersistenceException` with `cause` preserved. Same shape as `@kavo/typeorm`'s `mapDriverError`.
- [ ] `prisma-repository-adapter.ts`: implements `RepositoryAdapter` (`findOneById`, `findOne`, `findMany`, `count`, `create`, `update`, `patch`, `delete`, `restore`, `purge`) over `prismaClient[modelName]`. Soft delete supports the explicit-marker-field strategy (core's `context.config.softDelete`); Prisma has no `@DeleteDateColumn` equivalent, so zero-config detection via `softDeleteField` in metadata is not expected to apply the way it does for TypeORM — document this gap explicitly rather than faking it. Relation includes: core's `IncludeTree` join/batch distinction is collapsed onto Prisma's single `include` mechanism (a design note in the adapter should explain why batch-vs-join doesn't need to survive translation for Prisma, unlike TypeORM's query builder).
- [ ] `infrastructure.ts`: `createPrismaInfrastructure(prismaClient)` and `createPrismaKavo(prismaClient, options)`, mirroring `createTypeOrmInfrastructure`/`createTypeOrmKavo` exactly (per-entity metadata/adapter caching).
- [ ] `.dependency-cruiser.cjs`: new `prisma-only-imports-core` rule, same shape as `typeorm-only-imports-core` — `@kavo/prisma` may depend on `@kavo/core` and the `@prisma/client` peer only, never on `@kavo/nest`.
- [ ] Tests in `packages/orms/prisma/tests/` mirroring `@kavo/typeorm`'s `adapter.spec.ts`, `error-mapping.spec.ts`, `includes.spec.ts`, `soft-delete.spec.ts`, run against a real Prisma schema on SQLite for CI speed.
- [ ] `pnpm check` green (build, typecheck, depcruise, lint, test).
- [ ] Docs: `docs/` gets a short mention of Prisma as a second adapter wherever `@kavo/typeorm` is currently introduced as "the ORM adapter" (`docs/index.md`, `docs/README.md`, ADR-0011 consequences section already anticipates this — no ADR content change needed, just cross-reference).

## Affected packages

`@kavo/prisma` (new), `docs`. No changes to `@kavo/core`, `@kavo/typeorm`, or `@kavo/nest`.

## Constraints

- ADR-0001 / ADR-0005: core stays import-free; `@kavo/prisma` must only import the `@kavo/core` barrel (no deep imports) and the `@prisma/client` peer.
- ADR-0002: package topology — adapters never import each other or `@kavo/nest`; they meet only through Nest's DI container.
- ADR-0011: the metadata/infrastructure seam is the only ORM knowledge boundary — `@kavo/prisma` implements `CrudInfrastructure` and `RepositoryAdapter`, nothing else.
- ADR-0013: soft delete is config-declared, not adapter-invented — the adapter reports what it can detect via `softDeleteField`, core still decides strategy.
- ADR-0014: associate-by-id, not deep writes — relation writes follow the same id-association contract `@kavo/typeorm` follows.
- Naming (`CLAUDE.md`): `PrismaRepositoryAdapter`-style naming for the adapter class (named for what it adapts, matching `TypeOrmRepositoryAdapter`); `create*` factories.

## Out of scope

- `TransactionManager` — unimplemented even in `@kavo/typeorm` (no consumer yet); not part of this issue.
- Composite primary keys.
- Any change to `@kavo/core`, `@kavo/nest`, or `@kavo/graphql`.
- Migrating any existing example app off TypeORM — this is additive only.

h16d:
## What & why

Adds `@kavo/prisma`, a second `orms/*` adapter alongside `@kavo/typeorm`, implementing core's `RepositoryAdapter` and `CrudInfrastructure` seams over Prisma Client — full CRUD, filtering/sorting/pagination, explicit-field soft delete, and relation includes, with zero changes to `@kavo/core`.

The one real design departure from `@kavo/typeorm`: Prisma generates no runtime class for a model, so entities are identified by a caller-declared marker class matched to the schema by name, and a relation's target resolves through an explicit `entities` registry `createPrismaInfrastructure` takes. Written up in ADR-0017 since it's load-bearing, not just an implementation detail. Also notable: Prisma's `include` never multiplies root rows the way a raw SQL join does, so unlike `@kavo/typeorm` this adapter needs no join/batch translation for includes at all.

Closes #52

## Public API impact

New package only — `@kavo/core`, `@kavo/typeorm`, `@kavo/nest`, and `@kavo/graphql` are unchanged. `@kavo/prisma` exports `createPrismaInfrastructure`, `createPrismaKavo`, `PrismaRepositoryAdapter`, `buildEntityMetadata`, `translateFilter`, `mapDriverError`, and the supporting types.

## Testing

47 new tests across 4 spec files (`adapter`, `error-mapping`, `includes`, `soft-delete`), mirroring `@kavo/typeorm`'s suite, run against a real Prisma Client and schema on SQLite. `pnpm generate` (`prisma generate` + `prisma db push`) now runs first in `pnpm check`.

`pnpm check` (build, typecheck, depcruise, lint, test): **green** — 659 tests passing.

## Review notes

- `caseInsensitiveFilters` (default `true`) is a new, explicit option — Prisma's `mode: "insensitive"` filter isn't supported by every connector (SQLite, which this package's own tests run against, rejects it outright), and there's no reliable way to detect the connector from the DMMF at runtime.
- The associate-by-id write path (ADR-0014) requires the Prisma schema to declare the foreign key as an explicit scalar field; implicit many-to-many relations have no such field and can't be associated through the normal write path. Documented in the package README and architecture doc 14, not code-enforced.
- `TransactionManager` and composite primary keys are out of scope, matching `@kavo/typeorm`'s own posture.


h16d:
Ran the full reviewer fan-out (kavo-reviewer, kavo-test-auditor, kavo-security-auditor, kavo-perf-auditor). Security and perf came back clean. Two blocking findings, both fixed in 56882d5:

- **Boundary gap**: \`nest-only-imports-core\` / \`graphql-only-imports-core\` in \`.dependency-cruiser.cjs\` didn't include the \`@kavo/prisma\` package-specifier spelling, so \`@kavo/nest\`/\`@kavo/graphql\` could import it directly and depcruise wouldn't catch it (verified live with a probe import before/after the fix).
- **Test coverage**: the adapter's own distinguishing design surface — marker-class/entity-registry bootstrap errors (ADR-0017) and the \`caseInsensitiveFilters: true\` default branch — had zero coverage. Added \`tests/metadata.spec.ts\` and \`tests/filter-translator.spec.ts\`.

Also fixed two non-blocking findings: \`infrastructure.ts\`'s \`adapterFor\` now reuses \`metadataFor\`'s cache instead of re-deriving metadata, and \`docs/glossary.md\`'s Adapter entry now mentions \`PrismaRepositoryAdapter\`.

\`pnpm check\` green: 666 tests passing (one pre-existing, unrelated flake in \`@kavo/nest\`'s e2e suite, confirmed passing in isolation both before and after this branch's changes).

PR #97, opened 2026-08-03
test: bind e2e apps once instead of letting supertest bind per request

### LINKED ISSUE #91: Fix the ephemeral-port collision making the Nest/example e2e suites ~10% flaky
## Context

`packages/frameworks/nest/tests/binding.e2e.spec.ts` and the other Nest/example
e2e suites fail roughly 10% of the time locally, with a *different* test failing
each run. Measured at 12/117 runs on a feature branch **and** at the same rate on
pristine `main`, so it is pre-existing and environmental.

Root cause: `bootstrap()` calls only `await app.init()`, never `app.listen()`, so
`app.getHttpServer()` is a non-listening server. Every `request(server())` call
therefore makes supertest run `listen(0)` itself — on the **wildcard** — then
connect to a hardcoded `127.0.0.1` (supertest 7.2.2, `lib/test.js:60-70`). That
asymmetry is the bug: macOS will hand a wildcard `bind(0)` a port another process
already holds, and the connect to `127.0.0.1:<port>` is then delivered to that
other process. `binding.e2e.spec.ts` has 79 supertest calls, measured at ~130
`listen(0)`/`close` cycles per run — ~130 independent lotteries over the
ephemeral range.

Confirmed against `lsof`: `Spotify *:54090`, `agent-ser 127.0.0.1:59938`, plus
`Cursor`, `Ollama` and `rapportd` — 11 foreign ephemeral listeners on one
developer machine. The single mechanism explains every observed symptom:

| Symptom | Cause |
| --- | --- |
| `Parse Error: Expected HTTP/, RTSP/ or ICE/` | foreign non-HTTP greeting |
| a foreign `400` / `404` / `405` | a real status from someone else's HTTP server |
| `socket hang up` | peer accepts, then closes |
| hook (10s) or test (30s) timeout | peer accepts and never answers |

A foreign `405` is especially misleading — it reads as a missing generated route
or `KAVO_OPERATION_DISABLED`, sending you after a routing bug that does not exist.

Causality was isolated: re-running 30 times while refusing any ephemeral port
another local process holds in LISTEN state gave **0/30 failures** against a
15.6% baseline (p ≈ 0.006).

CI runners have almost no foreign ephemeral listeners, which is why this has not
been blocking CI — but `pnpm check` locally is the gate `/implement`, `/review`,
`/pr` and `/merge` all run, so a local green is currently ~90% reliable rather
than reproducible.

## Acceptance criteria

- [ ] `bootstrap()` binds once per app to the address supertest connects to
      (`await app.listen(0, "127.0.0.1")`) instead of letting supertest bind
      per-request
- [ ] every Nest and example e2e suite using the `app.init()` + supertest pattern
      is converted (six nest specs, incl. `settings.e2e.spec.ts`, plus the
      `examples/*` specs — `app.e2e.spec.ts` and `app-mariadb.e2e.spec.ts` both
      reproduced this)
- [ ] `binding.e2e.spec.ts` passes **30 consecutive** runs
- [ ] no test hardcodes or asserts on a port number
- [ ] `pnpm check` passes

## Affected packages

`@kavo/nest` (tests only), `examples/*` (tests only). No `src` changes.

## Constraints

- Tests stay in `tests/`, never `src/`.
- **Async-bind gotcha:** `listen(0, host)` resolves the host asynchronously, so
  `server.address()` is `null` immediately after the call — unlike the no-host
  path, which binds synchronously. The listen must be awaited before the port is
  read, or `request()` re-listens and the fix silently does nothing.
- Do not paper over it with retries, a fixed port, or a raised timeout; those
  hide the collision rather than removing it.

## Out of scope

- The two structural weaknesses in the same file worth doing while in
  `bootstrap()` but not required here: module-level `let app` / `let adapter`
  with a bare `afterEach(() => app.close())` that four tests must manually work
  around, and the five decoration-time-throw tests that never bootstrap and so
  re-close an already-closed app.

mahdavipanah:
## What and why

The Nest and example e2e suites failed roughly 10% of the time locally, with a different test failing each run — measured on pristine `main`, so it was pre-existing and environmental. `bootstrap()` called only `app.init()`, never `app.listen()`, leaving `getHttpServer()` unbound. Supertest therefore bound the server itself, once per request: `listen(0)` on the **wildcard**, then a connect to a hardcoded `127.0.0.1` (supertest 7.2.2, `lib/test.js:60-70`). That asymmetry is a lottery over the ephemeral range — macOS will hand a wildcard `bind(0)` a port another local process already holds, and the request is then delivered to that process. `binding.e2e.spec.ts` alone ran ~130 of those lotteries per run, which is why a foreign `405` (reading exactly like a missing generated route), a `Parse Error: Expected HTTP/`, a `socket hang up`, or a timeout could each show up.

A `listen(app)` helper now binds each app once — `await app.listen(0, "127.0.0.1")` — on the same address supertest connects to, and asserts the address is actually bound. That assertion is load-bearing: `listen(0, host)` resolves the host asynchronously, unlike the no-host path, so an unawaited bind would leave the server unbound and silently restore the old per-request behavior.

Six Nest specs and five example specs are converted. `base-kavo-mcp.controller.e2e.spec.ts` deliberately keeps `app.init()` — it resolves the toolset in process through DI and never makes an HTTP request, so it was never exposed to this.

Two structural guards keep it from coming back quietly. `binding.e2e.spec.ts` and `settings.e2e.spec.ts` serve requests from the server `bootstrap` listened on rather than re-deriving it from `app.getHttpServer()`, which answers just as happily for an unbound app. The examples' shared `crud-e2e.suite.ts` is parameterized and so cannot bind for itself; its `server()` throws with a pointed message if the registering spec did not.

The helper is duplicated across the three packages rather than shared: `.dependency-cruiser.cjs`'s `tests-no-other-package-internals` forbids a test file importing another package's `tests/`.

Closes #91

## Public API impact

None. Tests only — no `src` changes in any package, and no barrel changes.

## Testing

- New `packages/frameworks/nest/tests/listen.spec.ts` pins the helper's contract: the app is bound on the loopback before any request, and every request goes over that one binding instead of a new one each time. Verified failing-first — with `listen()` temporarily reverted to `app.init()`, both tests fail with "server is not bound to a TCP address".
- `binding.e2e.spec.ts` run **30 consecutive times: 0 failures**, against a ~10% baseline.
- `pnpm check` passes: build, typecheck, `depcruise` (247 modules, 885 dependencies, no violations), lint, and **898 tests across 55 files** — including the Testcontainers Postgres, MariaDB and MongoDB suites (Docker available locally).
- No test hardcodes or asserts a port number; the regression spec asserts the bound address is `127.0.0.1` and that the port is *unchanged* across requests, never a port value.

## Review notes

- Worth the hardest look: the duplicated `listen` helper. Three copies is the price of the boundary rule, and they will drift if someone edits one. If you would rather pay a different price (a shared dev-only workspace package, or relaxing the cruiser rule for a `tests/support` allowlist), this is the moment to say so.
- The `.claude/skills/write-tests/SKILL.md` addition is what stops the next e2e spec from reaching for `app.init()`. It also records the three non-fixes the issue rules out — a retry, a fixed port, a raised timeout.
- Deliberately left for later (the issue's own out-of-scope list): `binding.e2e.spec.ts`'s module-level `let app` / `let adapter` with a bare `afterEach(() => app.close())` that four tests work around by hand, and the five decoration-time-throw tests that never bootstrap and so re-close an already-closed app. I checked the interaction rather than assuming — `platform-express` 11.1.28's `close()` resolves rather than rejects on a non-running server, and the 30x run confirms it is harmless with a listening server.

### DIFF
--- a/.claude/skills/write-tests/SKILL.md
+++ b/.claude/skills/write-tests/SKILL.md
@@ -79,12 +79,19 @@ or a hook timeout, and it made these suites ~10% flaky (issue #91). A foreign
 `405` is the worst of them: it reads as a missing generated route.
 
 The `await` is load-bearing — `listen(0, host)` binds asynchronously, unlike
-the no-host path — so the helper asserts the address is bound: unreachable
-while the `await` is there, and it fires the moment someone drops it. The
-helper is duplicated per package (`packages/frameworks/nest/tests/support/`,
+the no-host path — so `listen` returns through `boundServer`, which rejects
+all three ways the binding can be wrong: absent, unbound, or bound to the
+wildcard. That last one is why a plain null-check is not enough — a wildcard
+bind reports a healthy-looking `::` address while supertest still connects to
+`127.0.0.1`. A suite that keeps its server in a variable routes its `server()`
+accessor through `boundServer` too, so a mid-test `close()` cannot slip past.
+
+The helper is duplicated per package (`packages/frameworks/nest/tests/support/`,
 `examples/*/tests/support/`) because a test file may not import another
-package's `tests/`; change all three copies together. Never fix a port
-collision with a retry, a fixed port, or a raised timeout; those hide it.
+package's `tests/` — the alternative, a shared home outside every package that
+owns it, is legal but buys less than it costs. Change all three copies
+together. Never fix a port collision with a retry, a fixed port, or a raised
+timeout; those hide it, and `listen.e2e.spec.ts` fails on the fixed port.
 
 ## Contracts that need wire-level assertions
 
--- a/.claude/skills/write-tests/SKILL.md
+++ b/.claude/skills/write-tests/SKILL.md
@@ -57,6 +57,34 @@ silently regressed:
 - **Regression** — a bugfix without a test that fails against the old code is
   unfinished. Write that test first and watch it fail.
 
+## HTTP e2e suites: bootstrap with `listen(app)`, never `app.init()`
+
+Any suite that drives a Nest app through supertest must bind it first:
+
+```ts
+import { listen } from "./support/listen.js"; // examples: "./listen.js"
+
+app = moduleRef.createNestApplication();
+const server = await listen(app); // == await app.listen(0, "127.0.0.1")
+await request(server).get("/todos").expect(200);
+```
+
+`app.init()` leaves `getHttpServer()` unbound, so supertest binds it itself,
+once per request: `listen(0)` on the **wildcard**, then a connect to a
+hardcoded `127.0.0.1`. That asymmetry is a lottery over the ephemeral range —
+a wildcard bind can be handed a port an unrelated local process already holds
+on the loopback, and the request goes to that process instead. It surfaces as
+`Parse Error: Expected HTTP/`, a foreign `400`/`404`/`405`, `socket hang up`,
+or a hook timeout, and it made these suites ~10% flaky (issue #91). A foreign
+`405` is the worst of them: it reads as a missing generated route.
+
+The `await` is load-bearing — `listen(0, host)` binds asynchronously, unlike
+the no-host path — so the helper asserts the address is bound. The helper is
+duplicated per package (`packages/frameworks/nest/tests/support/listen.ts`,
+`examples/*/tests/listen.ts`) because a test file may not import another
+package's `tests/`. Never fix a port collision with a retry, a fixed port, or
+a raised timeout; those hide it.
+
 ## Contracts that need wire-level assertions
 
 These cross a boundary to a consumer, so assert the actual shape:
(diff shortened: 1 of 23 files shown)


mahdavipanah:
CI note: the first run's `check (node lts/*)` job failed with a Prisma `P1008` socket timeout in `packages/orms/prisma/tests/soft-delete.spec.ts`, while `node current` and `node lts/-1` passed the identical commit. Re-running the job went green with no code change — all checks pass now.

It is unrelated to this PR (nothing here touches `@kavo/prisma`, the vitest config, or worker counts) and pre-existing: `@kavo/prisma`'s six spec files all share one SQLite file while vitest runs them as parallel workers, and three of them write via `deleteMany()` in `beforeEach`. SQLite admits one writer at a time, so on a slow/loaded runner the engine times out. Filed as #101 rather than folded in here.

PR #103, opened 2026-08-03
test(prisma): give every spec file its own SQLite database

### LINKED ISSUE #101: Fix the shared-SQLite contention making @kavo/prisma's suite flaky on CI
## Context

`check (node lts/*)` failed on PR #97 with a Prisma `P1008` while two other
Node versions passed the **identical commit**, and a re-run of the same job
went green. That is the signature of a load-dependent race, not a code defect —
and nothing in that PR touches `@kavo/prisma`, the vitest config, or worker
counts, so it is pre-existing and latent.

```
Invalid `client.invoice.deleteMany()` invocation in
  packages/orms/prisma/tests/soft-delete.spec.ts:54:24
Socket timeout (the database failed to respond to a query within the configured timeout).
Serialized Error: { code: 'P1008', meta: { modelName: 'Invoice' }, clientVersion: '6.19.3' }
```

Root cause: **one shared SQLite file, many parallel writers.**
`tests/support/client.ts` points every spec file at the same
`prisma/dev.db`, and vitest runs spec files in parallel worker threads. Three
of the six specs — `adapter`, `includes`, `soft-delete` — open `beforeEach`
with a chain of `deleteMany()` **writes**, so the write happens before *every*
test in each file, not once per file. SQLite admits one writer at a time; a
busy writer blocks the others, and Prisma's query engine gives up with `P1008`
once its socket timeout expires.

The models never overlap (`book`/`author` vs `note`/`article`/`blog` vs
`ticket`/`invoice`), so there is no logical data conflict — the contention is
purely the physical file lock. This is the same shape as #91: many independent
contention windows over a resource that was never meant to be shared, each one
a lottery. It does not reproduce on a fast local SSD (0 failures in 15
consecutive `pnpm vitest run packages/orms/prisma` runs); the window only opens
wide enough on a slower, loaded CI runner.

## Acceptance criteria

- [ ] `@kavo/prisma`'s spec files no longer contend for one SQLite file — each
      gets its own database, or the package's specs stop running as parallel
      writers
- [ ] whatever provides the per-file/per-worker database is wired into
      `pnpm generate` (or replaces the need for it), so `pnpm check` still works
      from a clean checkout with no manual setup
- [ ] `pnpm vitest run packages/orms/prisma` passes 30 consecutive runs
- [ ] `pnpm check` passes

## Affected packages

`@kavo/prisma` (tests only). Possibly `vitest.config.ts` depending on the
approach chosen. No `src` changes.

## Constraints

- Tests stay in `tests/`, never `src/`.
- Do not paper over it by raising Prisma's timeout, retrying, or adding
  `sleep`s — those hide the contention rather than removing it, exactly as
  ruled out for #91.
- Serializing the whole monorepo's test run to fix one package is too blunt;
  scope any parallelism change to `@kavo/prisma`.

## Out of scope

- The `beforeEach` write chains themselves. Deleting rows before every test is
  wasteful once each file owns its database, but that is a separate cleanup.

mahdavipanah:
## What

`@kavo/prisma`'s specs no longer share one SQLite file. `prisma db push` now
writes a **template** (`prisma/template.db`) that no test opens;
`tests/support/client.ts` copies it per client into a scratch directory, so the
spec files keep running in parallel workers with nothing contended between them.

## Why

`check (node lts/*)` failed on #97 with a Prisma `P1008` while two other Node
versions passed the identical commit, and a re-run went green — a load-dependent
race, not a code defect. Every spec pointed at `prisma/dev.db`, vitest runs spec
files in parallel workers, and three of the six specs open `beforeEach` with a
chain of `deleteMany()` **writes**. SQLite admits one writer at a time, so a busy
writer could hold the file lock long enough for another file's query to exhaust
the query engine's socket timeout. The models never overlap — the contention was
purely the physical file lock.

Removing the shared resource, rather than widening the timeout around it, is the
fix (the same disposition as #91).

Two details worth review attention:

- **Cleanup runs in the main process.** My first attempt cleaned up via
  `process.on("exit")` inside the worker and leaked 8 directories per run —
  vitest kills its workers, so worker-side exit handlers never fire. A
  `globalSetup` provisions one scratch root per run and removes it at teardown.
  Per run rather than at a fixed path, so two concurrent vitest runs cannot
  delete each other's databases.
- **Recovery sidecars are copied** (`-journal`, `-wal`) when present, so the copy
  carries every committed transaction the template does. `-shm` is deliberately
  not copied — see the review commit below.

A missing template now fails with ``run `pnpm generate` first`` instead of
silently creating an empty database that fails later with "no such table".

## Public API impact

None. Tests, test support, `vitest.config.ts`, `.gitignore` and docs only — no
`src` changes, no change to any published package's surface. `pnpm generate` is
otherwise unchanged, so a clean checkout still needs no manual setup.

## Testing

- **`pnpm vitest run packages/orms/prisma` × 30 consecutive: all green**, with 0
  leaked scratch directories afterwards. Re-run in full after each round of
  review fixes.
- **6 concurrent full prisma runs: all 6 exit 0.** This is the check that
  actually exercises the fix — sequential local runs never reproduced the
  failure (0/15 on `main`, per the issue), while concurrent runs put real
  contention on the shared file.
- **`tests/database-isolation.spec.ts`** pins the whole contract: one database
  per client, copies land inside the run's scratch root byte-for-byte, the
  scratch root is a directory during the run and gone after teardown, the
  `.env` ↔ `TEMPLATE_DATABASE` filename pair agrees, the missing-template and
  unset-scratch-root guards fire, and a copy carries committed-but-uncheckpointed
  WAL frames without the stale wal-index.
- **Verified by mutation, not by inspection.** Six separate breaks — sharing one
  database, dropping `-wal`, re-adding `-shm`, provisioning outside the scratch
  root, a teardown that stops deleting, and a diverging `.env` filename — each
  killed by exactly one test, with the baseline green.
- **`pnpm check` passes**: 58 files, 950 tests, depcruise clean (252 modules,
  897 dependencies), typecheck and lint clean. `pnpm docs:links` and
  `pnpm format:check` pass.

## Review notes

Out of scope per the issue: the `beforeEach` `deleteMany()` chains are untouched.
They are now redundant work rather than a contention source — a separate cleanup.

`/code-review` raised two findings, both fixed in the third commit:

- Copying the `-shm` sidecar is unsafe. It is a derived wal-index SQLite rebuilds
  on open; if the template ever carries a `-wal`, a stale copied `-shm` can make
  SQLite skip recovering committed frames, so the copy silently loses rows.
- Falling back to `os.tmpdir()` when `KAVO_PRISMA_SCRATCH_ROOT` is unset degraded
  silently — a process where `globalSetup` did not run would pass while leaking a
  fixture database per client. It now throws, naming the variable and the file
  that sets it.

Docs resynced: `CONTRIBUTING.md`'s setup section and
`docs/internals/architecture/14-prisma-adapter.md` §7 both described `pnpm
generate` as pushing the database the tests read.

Closes #101

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ api.rest
 docs/.vitepress/cache
 packages/orms/prisma/prisma/*.db
 packages/orms/prisma/prisma/*.db-journal
+packages/orms/prisma/prisma/*.db-wal
+packages/orms/prisma/prisma/*.db-shm
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,6 +51,13 @@ pnpm build
 fixture schema is SQLite, and `packages/orms/prisma/prisma/.env` is committed
 with a working `DATABASE_URL`.
 
+What it pushes is a **template**, `prisma/template.db`, which no test opens.
+`tests/support/client.ts` copies it per client, so each spec file writes to a
+database of its own — SQLite admits one writer at a time, and one shared file
+made the suite flaky under CI load. The copies live in a scratch directory
+that `tests/support/global-setup.ts` creates per vitest run and deletes when
+the run ends.
+
 Note what it is _not_ for. `@kavo/prisma`'s own `src` never imports
 `@prisma/client` — it models the client structurally in
 `prisma-client-like.ts`, precisely so a library package does not assume the
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -167,6 +167,13 @@ A few things about the test setup that are easy to trip over:
 - **The SWC transform is required.** TypeORM entities and Nest DI rely on
   decorator metadata, which vitest's default esbuild transform cannot emit. That
   is why `unplugin-swc` is in the vitest config — don't remove it.
+- **`@kavo/prisma`'s specs need the `globalSetup` in the vitest config.** It
+  creates the scratch directory their database copies are written into and
+  deletes it when the run ends. Remove it, or run those specs under a config
+  that omits it, and every `newTestPrismaClient()` throws
+  `KAVO_PRISMA_SCRATCH_ROOT is unset` — the fixture provisioner refuses to fall
+  back to the OS temp directory, because a copy made outside that root is one
+  nothing ever deletes.
 - **An e2e spec binds its app with `listen(app)`, never `app.init()`.** The
   helper lives in each package's `tests/support/listen.ts`. `init()` leaves
   `getHttpServer()` unbound, so supertest binds it per request — `listen(0)` on
--- a/docs/internals/architecture/14-prisma-adapter.md
+++ b/docs/internals/architecture/14-prisma-adapter.md
@@ -114,3 +114,12 @@ SQLite (`tests/adapter.spec.ts`, `tests/soft-delete.spec.ts`,
 `tests/includes.spec.ts`), per-package testing as specified — the shared
 test schema and generated client are built by `pnpm generate`
 (`prisma generate` + `prisma db push`), which `pnpm check` runs first.
+
+The schema is shared; the database is not. `db push` writes a template
+(`prisma/template.db`) that no test opens, and `tests/support/client.ts`
+copies it per client into a scratch directory `tests/support/global-setup.ts`
+provisions for the run and removes afterwards. Vitest runs spec files in
+parallel workers and SQLite admits one writer at a time, so a single file
+turned each `beforeEach` write chain into a lock other files could stall
+behind — a `P1008` socket timeout under CI load. One database per client
+removes the contention instead of widening the timeout around it.
--- a/packages/orms/prisma/prisma/.env
+++ b/packages/orms/prisma/prisma/.env
@@ -1,1 +1,1 @@
-DATABASE_URL="file:./dev.db"
+DATABASE_URL="file:./template.db"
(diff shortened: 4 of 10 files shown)


mahdavipanah:
Follow-up: the review fan-out found no correctness defects, but three invariants this change's own comments assert were unpinned — each breakable while the whole suite stayed green. All are now covered, and the coverage is verified by mutation rather than by inspection: six separate breaks (sharing one database, dropping `-wal`, re-adding `-shm`, provisioning outside the scratch root, a teardown that stops deleting, a diverging `.env` filename) are each killed by exactly one test, baseline green.

The sidecar rule got a behavioral test rather than a fake-bytes one: it builds a database with a genuinely uncheckpointed WAL — a holder client that *reads* takes SQLite's deadman-switch lock, so the writer's close skips its checkpoint — then asserts the committed row survives the copy while no wal-index tags along. A guard asserts the WAL is really there, so a future SQLite that checkpoints on close fails loudly instead of passing vacuously. `-journal` is left uncovered on purpose, with the reason in the comment.

Also: `vi.stubEnv` replaces the hand-rolled `process.env` save/restore, the spec has its own `Scratch` fixture model instead of borrowing `Invoice` from soft-delete's section, and `metadata.spec.ts` disconnects its client (it now holds a scratch database open at teardown — harmless on POSIX, `EBUSY` on Windows).

PR #104, opened 2026-08-03
feat(mikroorm): add @kavo/mikroorm adapter with parity to @kavo/typeorm

### LINKED ISSUE #102: Add @kavo/mikroorm adapter with feature parity to @kavo/typeorm
## Context

Kavo currently ships three data-access adapters (`@kavo/typeorm`, `@kavo/prisma`, `@kavo/mongoose`), each implementing core's `RepositoryAdapter` seam per ADR-0002's `packages/orms/*` topology. MikroORM is a common alternative to TypeORM with a similar decorator-based (`@Entity()`) entity model, and adding it extends Kavo's ORM coverage without touching core. Since MikroORM entity classes are real runtime classes carrying their own metadata (via MikroORM's `MetadataStorage`), this adapter should mirror `@kavo/typeorm`'s approach directly — reading decorator metadata at bootstrap — rather than the marker-class indirection `@kavo/prisma` needs (ADR-0017), which exists specifically because Prisma erases its models at compile time.

## Acceptance criteria

- [ ] New package `packages/orms/mikroorm` (`@kavo/mikroorm`), following the `packages/orms/<name>` topology (ADR-0002), with `mikro-orm`/`@mikro-orm/core` as a peer dependency and `@kavo/core` as the only workspace dependency.
- [ ] Implements core's `RepositoryAdapter` (`EntityReader` + `EntityWriter`) over a MikroORM `EntityManager`/`MikroORM` instance, named `MikroOrmRepositoryAdapter` per the adapter-naming convention.
- [ ] `buildEntityMetadata` feeds core's entity-metadata seam from MikroORM's own metadata (id field, scalar fields with type/nullability, relation descriptors, soft-delete marker column detection where MikroORM exposes one) — same shape/contract as `@kavo/typeorm`'s `buildEntityMetadata`.
- [ ] `FilterTranslator` maps the Filter AST (`SCREAMING_SNAKE` operators) to MikroORM's query/`QueryBuilder` API, matching `@kavo/typeorm`'s filter-translator scope (AND/OR/NOT groups, relation-path filters).
- [ ] `mapDriverError` maps MikroORM/underlying-driver errors to Kavo's exception hierarchy, matching `@kavo/typeorm`'s error-mapping scope.
- [ ] `createInfrastructure` / `createMikroOrmKavo` sugar analogous to `@kavo/typeorm`'s, deriving adapter + metadata from a MikroORM instance.
- [ ] Full feature parity with `@kavo/typeorm`'s adapter-level scope: filtering, sorting, pagination, optional counting, nested includes/relation loading (doc 12), soft delete/restore/purge (doc 11), and transactions.
- [ ] Test suite under `packages/orms/mikroorm/tests` mirroring `@kavo/typeorm`'s coverage: adapter CRUD/query behavior, filter-translator unit tests, error-mapping, includes, soft-delete — using the same Vitest+SWC setup as the other adapters (real SQLite/driver, not mocks, matching existing adapter test conventions).
- [ ] New architecture doc `docs/internals/architecture/16-mikroorm-adapter.md` (next available number) documenting the metadata seam, query translation, and any adapter-specific caveats, mirroring docs 09/14/15.
- [ ] Integration doc under `docs/integrations/nest/mikroorm.md` mirroring the existing per-ORM wiring docs.
- [ ] `.dependency-cruiser.cjs` gets a `mikroorm-only-imports-core` rule matching the existing `typeorm-only-imports-core` / `prisma-only-imports-core` / `mongoose-only-imports-core` pattern.
- [ ] `pnpm check` passes (build, typecheck, depcruise, lint, test) with the new package included.

## Affected packages

`@kavo/mikroorm` (new), `@kavo/core` (none expected — read-only against existing seams), docs, `.dependency-cruiser.cjs`, `pnpm-workspace.yaml` (already globs `orms/*`, no change needed).

## Constraints

- **ADR-0002** (package topology): lives under `packages/orms/mikroorm`; `orms/*` workspace glob already covers it.
- **ADR-0005** (core purity): zero changes to `@kavo/core`'s zero-runtime-dependency guarantee; the adapter imports only the `@kavo/core` barrel, never deep imports.
- Package-boundary rule: `@kavo/mikroorm` never imports `@kavo/nest` or any other adapter — enforced by the new depcruise rule.
- Naming conventions (CLAUDE.md): `MikroOrmRepositoryAdapter`, `create*` factories (`createMikroOrmKavo`), no `I`-prefixed interfaces, `*Exception` classes reused from core (no new exception types expected).
- Tests live in `tests/`, never `src/`, per repo convention.

## Out of scope

- Any changes to `@kavo/core`'s public contracts or barrel beyond what's already stable for existing adapters.
- GraphQL/MCP protocol binding work — those already delegate through core's engine and need no per-ORM changes.
- New Kavo operations or config keys — this issue is adapter parity only, not new CRUD capability.

h16d:
## What

Adds `@kavo/mikroorm` (`packages/orms/mikroorm`), the fourth `orms/*` adapter: `RepositoryAdapter` over a MikroORM `EntityManager`, with the metadata seam fed from MikroORM's own `MetadataStorage`.

Closes #102

## Why this shape

MikroORM sits between the two existing SQL adapters — TypeORM's entity model (real decorated runtime classes carrying their own metadata) with Prisma's query surface (a declarative `FilterQuery` that nests relation paths natively). So the metadata seam mirrors `@kavo/typeorm` and the translation seam mirrors `@kavo/prisma`. No marker classes (ADR-0017), and no ADR of its own: the class passed to `createCrud` is the class MikroORM registered.

Four adapter-specific decisions, all documented in [doc 17](docs/internals/architecture/17-mikroorm-adapter.md):

- **Every operation forks its own `EntityManager`.** MikroORM is a Unit-of-Work ORM whose EM owns an identity map; sharing one across requests would serve stale rows. That is why `createInfrastructure` takes the `MikroORM` instance rather than an EM — it needs something to fork from.
- **Rows convert with `wrap().toObject()` at the boundary.** Required, not cosmetic: a to-many is a `Collection`, not an array, and core's `DefaultSerializer` branches on `Array.isArray` — a `Collection` would fall down the single-row path and serialize its internals. `@kavo/mongoose` converts at the same seam for the same reason.
- **`IncludeNode.strategy` is ignored**, as in `@kavo/prisma`. MikroORM resolves `populate` with its own queries and applies `limit`/`offset` to the root either way, so the join/batch split has nothing to control here.
- **`softDeleteField` is always `null`.** MikroORM's soft-delete pattern is a user-defined `@Filter` — a query concern, not a column declaration — so `softDelete.field` must be configured.

## Two calls worth a reviewer's eye

**`caseInsensitiveFilters` defaults to `false`**, where `@kavo/prisma`'s identically-named option defaults to `true`. `$ilike` is PostgreSQL-only and every other driver takes the token verbatim and fails with a syntax error, so the failure modes are not symmetric: off on Postgres costs case-insensitivity, on anywhere else makes *every* `ILIKE` query throw. On SQLite it isn't even a loss — its `LIKE` is already ASCII case-insensitive.

**Empty groups are spelled as an empty `$in` on the primary key.** MikroORM renders `$not: {}` as match-*everything*, the opposite of what an empty `NOT` group means, and rejects an empty `$and`/`$or` outright. An empty `$in` is the one portable contradiction. Verified against the real driver, not assumed.

## Public API impact

New package barrel only — `@kavo/core`'s barrel is untouched, and core has no changes at all. Exports: `MikroOrmRepositoryAdapter`, `buildEntityMetadata`, `mapDriverError`, `translateFilter` + `MikroWhere` + `FilterTranslatorOptions`, `createInfrastructure` / `createMikroOrmKavo` / `MikroOrmInfrastructureOptions`, and `toPlain` / `toPlainAll`.

That last pair is a deliberate call rather than an oversight: it has no consumer outside the package today, but it is the documented boundary conversion and belongs alongside the other seams an adopter may need to reimplement. Say the word if you'd rather it stayed internal.

## Deviations from the issue

- **Doc is `17-`, not `16-`** — `16-mcp-binding.md` already exists; the issue said "next available number."
- **Peer is `^6.0.0`, not `^6 || ^7`.** MikroORM v7 removed decorators entirely for `defineEntity`/`EntitySchema`, invalidating the premise the entity-identity design rests on. Not claimed until tested.
- **No transaction plumbing.** Parity here *is* absence: `@kavo/typeorm` never reads `context.transaction` either, and core's `TransactionManager` has no consumer. Building one would exceed "adapter parity only."

## Testing

`pnpm check` green: **63 files, 1054 tests**. `pnpm docs:links`, `pnpm docs:build`, `pnpm format:check` all green.

91 new tests across 5 spec files, on real SQLite rather than mocks, matching the other adapters' conventions: adapter CRUD/query behaviour, filter-translator units, error mapping, includes, soft delete, metadata derivation.

Two of them exist because the implementation found real problems:

- **Embeddable child properties leaked.** MikroORM exposes both the addressable parent and one child per inner column (`address~city`, `billing_city`); unfiltered they land in derived DTOs and allowlists as unusable wire fields. Now dropped, with a test.
- **Relation deep-write.** Core only narrows a relation to `{ id }` when the *target* is in its catalog, so a relation whose target was never `createCrud`-ed arrives as whatever the client sent. An object carrying no id now becomes `null` rather than reaching `em.create`, which would leave a nested write one cascade setting away from working (ADR-0014). Regression test included.

## Review notes

- **The depcruise change is more than the one new rule.** Besides `mikroorm-only-imports-core`, `mikroorm` joins the four `to:` alternations that guard *other* packages against importing an adapter. Without those, a bare `@kavo/mikroorm` specifier from core or a protocol would pass a green `depcruise` — exactly the gap CLAUDE.md warns about.
- **One tooling change.** `@mikro-orm/better-sqlite@6` pins `better-sqlite3@^11`, whose prebuilds stop before current Node, so install fails outright. The root manifest carries a **scoped** `pnpm.overrides` entry (`"@mikro-orm/better-sqlite>better-sqlite3"`) raising that dependency alone — `@kavo/typeorm`'s own SQLite resolution is untouched. Explained in CONTRIBUTING.
- **The docs sweep is deliberately wider than the issue asked.** Docs 01/02/06/11/12 carried claims an eighth package falsifies — notably "only `@kavo/typeorm` acts on the join/batch distinction" — and CLAUDE.md/CONTRIBUTING.md counted seven packages. `publish.yml` and `/publish`'s table also needed the new entry, which `tests/release-workflow.spec.ts` gates against the real workspace.
- **`ToolLogoStrip.vue` is untouched.** The homepage logo strip needs an SVG path I don't have, and fabricating a logo is worse than omitting it.

### DIFF
--- a/.claude/commands/publish.md
+++ b/.claude/commands/publish.md
@@ -48,7 +48,7 @@ workflow.
 3. **Apply the version, in lockstep** (ADR-0004 — [`docs/internals/adr/0004-lockstep-versioning.md`](../../docs/internals/adr/0004-lockstep-versioning.md)):
    set the new version in the `package.json` of **every** published package.
    `PACKAGE_DIRS` in `.github/workflows/publish.yml` is the single source of
-   truth for that set — read it and bump exactly those. Today it is all seven:
+   truth for that set — read it and bump exactly those. Today it is all eight:
 
    | Directory                    | Package          |
    | ---------------------------- | ---------------- |
@@ -56,6 +56,7 @@ workflow.
    | `packages/orms/typeorm`      | `@kavo/typeorm`  |
    | `packages/orms/prisma`       | `@kavo/prisma`   |
    | `packages/orms/mongoose`     | `@kavo/mongoose` |
+   | `packages/orms/mikroorm`     | `@kavo/mikroorm` |
    | `packages/protocols/graphql` | `@kavo/graphql`  |
    | `packages/protocols/mcp`     | `@kavo/mcp`      |
    | `packages/frameworks/nest`   | `@kavo/nest`     |
--- a/.dependency-cruiser.cjs
+++ b/.dependency-cruiser.cjs
@@ -8,6 +8,7 @@
  *      │                ▲
  *      │                ├── @kavo/prisma
  *      │                ├── @kavo/mongoose
+ *      │                ├── @kavo/mikroorm
  *      │                ├── @kavo/graphql ◀─┐
  *      │                └── @kavo/mcp     ◀─┤
  *      └── the one sanctioned sideways edge ┘
@@ -73,6 +74,18 @@ module.exports = {
       to: { path: "^(packages/frameworks|@kavo/nest)" },
     },
     {
+      name: "mikroorm-only-imports-core",
+      severity: "error",
+      comment:
+        "@kavo/mikroorm may depend on @kavo/core and the @mikro-orm/core " +
+        "peer — never on @kavo/nest (ADR-0002). Both spellings are matched: " +
+        "a workspace package specifier does not resolve to a path here, so a " +
+        'path-only rule would miss `from "@kavo/nest"` — the spelling ' +
+        "anyone would actually write.",
+      from: { path: "^packages/orms/mikroorm/src" },
+      to: { path: "^(packages/frameworks|@kavo/nest)" },
+    },
+    {
       name: "nest-only-imports-core",
       severity: "error",
       comment:
@@ -85,7 +98,7 @@ module.exports = {
         "@kavo/nest back (ADR-0016). Package-specifier form matched too, " +
         "per the note on typeorm-only-imports-core.",
       from: { path: "^packages/frameworks/nest/src" },
-      to: { path: "^(packages/orms|@kavo/(typeorm|prisma|mongoose))" },
+      to: { path: "^(packages/orms|@kavo/(typeorm|prisma|mongoose|mikroorm))" },
     },
     {
       name: "graphql-only-imports-core",
@@ -97,7 +110,7 @@ module.exports = {
         "same constraint as an ORM adapter). Package-specifier form matched " +
         "too, per the note on typeorm-only-imports-core.",
       from: { path: "^packages/protocols/graphql/src" },
-      to: { path: "^(packages/orms|packages/frameworks|@kavo/(typeorm|prisma|mongoose|nest))" },
+      to: { path: "^(packages/orms|packages/frameworks|@kavo/(typeorm|prisma|mongoose|mikroorm|nest))" },
     },
     {
       name: "mcp-only-imports-core",
@@ -110,7 +123,7 @@ module.exports = {
         "Package-specifier form matched too, per the note on " +
         "typeorm-only-imports-core.",
       from: { path: "^packages/protocols/mcp/src" },
-      to: { path: "^(packages/orms|packages/frameworks|@kavo/(typeorm|prisma|mongoose|nest))" },
+      to: { path: "^(packages/orms|packages/frameworks|@kavo/(typeorm|prisma|mongoose|mikroorm|nest))" },
     },
     {
       name: "no-cross-package-deep-imports-core",
@@ -161,7 +174,7 @@ module.exports = {
         "core's tests may use the barrel and vitest; this keeps the part that " +
         "still matters — no adapter, no framework — enforced for them too.",
       from: { path: "^packages/core/tests" },
-      to: { path: "^(@kavo/(typeorm|prisma|mongoose|nest|graphql|mcp)|packages/(orms|frameworks|protocols))" },
+      to: { path: "^(@kavo/(typeorm|prisma|mongoose|mikroorm|nest|graphql|mcp)|packages/(orms|frameworks|protocols))" },
     },
     {
       name: "no-circular",
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -28,6 +28,7 @@ jobs:
         packages/orms/typeorm
         packages/orms/prisma
         packages/orms/mongoose
+        packages/orms/mikroorm
         packages/protocols/graphql
         packages/protocols/mcp
         packages/frameworks/nest
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -4,7 +4,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 ## What Kavo is
 
-A production-grade CRUD framework for TypeScript: define an entity once (via TypeORM, Prisma, or Mongoose) and get the full REST CRUD surface — filtering, sorting, pagination, nested includes, field selection, optional per-operation DTOs, transactions, and problem-details errors — behind generated NestJS routes, configurable at global → entity → operation → per-call scope.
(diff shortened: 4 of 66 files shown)


h16d:
Pushed a follow-up commit after a full reviewer fan-out (correctness, tests, security) plus building out `examples/nest-mikroorm` — the same Pet domain `nest-typeorm` serves, on this adapter.

## What the second commit fixes

**Correctness**
- `findOne(query)` with no filter threw a 500. MikroORM's validator rejects `em.findOne` with an empty `where`; `em.find` accepts it, and every sibling adapter returns the first row for this shape. Now routed through `em.find(..., { limit: 1 })`.
- A relation declared by name (`@ManyToOne("Owner")` — the spelling that keeps a bidirectional relation's cycle off the runtime import graph, which this repo's own `no-circular` depcruise rule requires) left `property.entity` a **string**, not a thunk. Calling it threw. The Pet domain's `Owner`↔`Pet` cycle is exactly what surfaced this; the original suite's schema never needed the string form. Resolved via `targetMeta` now.
- The single-table-inheritance discriminator (`species` on Cat/Dog) landed in the *writable* projection — a client could set `species: "cat"` on a Dog and write a row its own repository could no longer load. Now reported generated.
- **The one I'd flag hardest for review**: soft-deleted rows survived a *nested* include (`?include=articles.notes`) by dropping the **parent** row entirely instead of returning an empty child array. MikroORM's `populateWhere` reads a nested condition as a relation-path predicate on the parent — I probed three spellings (nested, dotted, parent-only), all wrong in a different way, so the adapter now populates everything and prunes soft-deleted rows in memory at any depth, guaranteeing every included relation is present (`[]`/`null`) rather than missing.

**Security** (verified, not just alleged)
- Doc 17 previously said "soft delete needs an explicit `softDelete.field`" — false. `softDelete` defaults to `{ field: "deletedAt", strategy: "auto" }` and core matches by column **name**, so any entity with a plain `deletedAt` property is soft-deletable with zero config, while the marker stays client-writable (MikroORM has nothing to mark it generated with). Corrected everywhere and pinned with tests.
- A `hidden`/`lazy` MikroORM property was droppable from responses but stayed on the default filter/sort allowlists — a blind extraction oracle (`filter[passwordHash][like]=a%` answered by row count even though the value never appears in a body). Now excluded from the metadata seam entirely, same posture `@kavo/mongoose` already takes for `select: false`.
- Documented, not fixed (parity with `@kavo/typeorm`'s identical `@PrimaryColumn` gap, not a regression): a non-auto-increment primary key (`@PrimaryKey() id: string = v4()`) is client-writable.

## The new example

`examples/nest-mikroorm`: Owner/Cat/Dog/Tag/Address, single-table inheritance, relation-path filtering/sorting, config-declared soft delete — over in-memory SQLite by default and a Testcontainers Postgres for the one thing that needs it (`caseInsensitiveFilters: true`, since `$ilike` is PostgreSQL-only).

## Gate

`pnpm check`: **65 files, 1128 tests**. `docs:links`, `docs:build`, `format:check` all green.

Issue #118, opened 2026-08-04
Add cursor-based pagination strategy


h16d:
## Context

Kavo currently ships two pagination strategies (`offset`, `page` — `packages/core/src/query/pagination-strategies.ts`), both normalizing to the same internal `{ limit, offset }` form (`packages/core/src/query/pagination.ts`), which every adapter consumes via `skip`/`take` (or the Prisma/Mongoose/MikroORM equivalents) directly. Offset pagination degrades on large tables and is unstable under concurrent writes (skipped/duplicated rows as data shifts between pages). A cursor (keyset) strategy is the standard fix and is a natural extension of the existing pluggable `PaginationStrategy` seam — `PaginationStrategyName` is already typed as open (`"offset" | "page" | (string & {})`).

## Acceptance criteria

- [ ] A built-in `CursorPaginationStrategy` registered under `pagination.strategy: "cursor"` in `builtInPaginationStrategies()`.
- [ ] The internal `Pagination` form represents cursor semantics (opaque cursor + limit) distinctly from offset semantics, so an adapter never receives a meaningless `offset` for a cursor query.
- [ ] Cursor value is opaque to clients (encodes the last row's sort-key values) and tamper-resistant; a malformed/invalid cursor is rejected with a query validation error, consistent with how `page[number]`/`limit`/`offset` are validated today.
- [ ] Cursor pagination requires the effective `sort` to end in a unique tiebreaker field (e.g. `id`); requests using `pagination.strategy: "cursor"` without one are rejected at query-normalization time.
- [ ] `EntityReader`/adapters for `@kavo/typeorm`, `@kavo/prisma`, `@kavo/mongoose`, and `@kavo/mikroorm` translate a cursor into a keyset `WHERE`/filter clause plus `ORDER BY`, replacing `skip`, while still honoring existing filter, include, and soft-delete behavior.
- [ ] `ListResultDto`/`ListMetaDto` reports the next cursor (e.g. `meta.nextCursor`, `null` on the last page) without adding new top-level envelope fields.
- [ ] Tests (per touched package) cover: first/middle/last page, empty result set, invalid/tampered cursor, cursor combined with filters and includes, and rejection when no unique sort tiebreaker is present.
- [ ] `pnpm check` passes.
- [ ] Docs (`docs/using-the-api.md` and/or per-ORM integration docs) document `pagination.strategy: "cursor"`, cursor opacity, and the tiebreaker requirement.

## Affected packages

[...]

PR #126, opened 2026-08-04
feat(core): populate ListResultDto.meta from the findMany handler

### LINKED ISSUE #122: Populate `ListResultDto.meta` from the `findMany` handler instead of discarding it
## Context

`ListResultDto.meta` (`packages/core/src/dto/list-result.ts`) is already documented as "the extension seam for pagination strategies (`meta.cursor`) and consumer middleware" — but nothing actually populates it. `KavoEngine.mapResponse` (`packages/core/src/engine/kavo-engine.ts`) hardcodes `meta: {}` on every `findMany` response, and `FindManyResult<Entity>` (`packages/core/src/engine/built-in-handlers.ts`) — what the `findMany` handler returns — has no `meta` field for a handler to even attach one to. `builtInHandlers()` is already a public export, so an adopter can already wrap the built-in `findMany` handler via `operations.findMany.handler` to add custom logic; the only thing missing is a place to put the result and an engine that doesn't throw it away. This is a small, surgical fix that turns an already-documented seam into a real one — and it's also the exact mechanism cursor pagination (#118, `meta.nextCursor`) will need, so this issue lands the seam both features share.

## Acceptance criteria

- [ ] `FindManyResult<Entity>` gains an optional `meta?: Readonly<Record<string, unknown>>`.
- [ ] `KavoEngine.mapResponse` uses `result.meta ?? {}` for the envelope's `meta` field instead of the hardcoded `{}`.
- [ ] A custom `findMany` handler (via `operations.findMany.handler`, wrapping `builtInHandlers(adapter).findMany` or the entity's configured handler) can return `meta` and see it verbatim on `ListResultDto.meta` — no new config key required, since the override mechanism already exists.
- [ ] A small ergonomic helper (e.g. `withListMeta(handler, compute)` in `@kavo/core`) that wraps any `findMany`-shaped `OperationHandler` and merges `compute(result, context)`'s return value into `meta`, so the common "just add one stat" case doesn't require hand-writing the full wrap-and-spread boilerplate every time.
- [ ] Tests cover: default `findMany` still returns `meta: {}` when no handler sets it; a wrapped handler's `meta` reaches the wire response untouched; the `withListMeta` helper merges correctly and doesn't clobber a `meta` key a caller intentionally set; `meta` values survive serialization (arbitrary JSON-serializable shapes, not just primitives).
- [ ] `pnpm check` passes.
- [ ] `docs/using-the-api.md` (or wherever custom operation handlers are documented) gets a short section showing how to add custom list-response metadata via a wrapped `findMany` handler.

## Affected packages

`@kavo/core`, docs — no ORM adapter, `@kavo/nest`, or protocol-binding changes; `meta` already flows through the existing envelope to every transport unchanged.

## Constraints

- ADR-0006 (registry-driven operations): this is wiring, not a new special case — `findMany` already supports handler overrides like every other operation; the fix is that the engine stops discarding part of what the handler returns.
- `ListMetaDto` stays an open bag (`docs/internals/architecture/04-dto-system.md`'s envelope-fields convention: `items`, `limit`, `offset`, `total`, `meta`) — this issue does not add new first-class envelope fields, only makes the existing `meta` field actually reachable.
- Don't confuse this with `OperationConfig.meta` (`OperationMetadata`, ADR-0007) — that's route/framework metadata on the registry entry, a completely different concept from `ListResultDto.meta` on the response envelope. Naming in the new helper/docs should make the distinction unambiguous (e.g. never call the new hook just "meta").

## Out of scope

- Item (`findOne`/`createOne`/etc.) responses gaining a `meta` field — only `ListResultDto` has one; single-item responses have no envelope to extend.
- Building cursor pagination's `meta.nextCursor` itself — that's #118's job; this issue only lands the generic seam it (and everything else) will use.
- A declarative (non-function) way to configure static/computed meta per entity — this issue only unblocks the handler-return-value path; a richer declarative API can be a follow-up if the wrap-and-return pattern proves too manual in practice.

mahdavipanah:
## What & why

`ListResultDto.meta` was documented as "the extension seam for pagination strategies and consumer middleware", but nothing could actually reach it: `KavoEngine.mapResponse` hardcoded `meta: {}` and destructured only `{ entities, total }` off the handler result, so anything else a `findMany` handler returned was silently discarded. This turns that documented seam into a real one — `FindManyResult` gains an optional `meta`, the engine merges it into the envelope, and `withListMeta(handler, compute)` wraps any `findMany` handler so the common "just add one stat" case doesn't need hand-written wrap-and-spread. No new config key: `operations.findMany.handler` already exists (ADR-0006), so this is wiring, not a new special case. It is also the exact mechanism cursor pagination (#118) will need for `meta.nextCursor`, which is why the engine's merge point is a named step (`listMeta`) rather than an inline `?? {}` — an engine-side contributor folds in there without every handler having to know about it.

Following review, **`meta` is also now optional on the wire**: the engine leaves the key off the envelope entirely unless something contributed, rather than shipping `"meta": {}` on every list. See "Public API impact" below — this is the one breaking part of the PR.

Closes #122

## Public API impact

- **`FindManyResult.meta`** — a new **optional** field on a public type, additive and non-breaking. Every handler written before it still satisfies the contract, and the built-in handler never sets it.
- **`ListResultDto.meta` is now optional (`meta?: ListMetaDto`) and omitted when empty** — **breaking**. `@kavo/core@0.6.0` ships `meta: {}` on every list response (`mapResponse` hardcoded it on `main`, so this predates the PR). A consumer reading `result.meta.x` must now read `result.meta?.x`. Rationale in "Review notes".
- **Two new named barrel exports** in `packages/core/src/index.ts` — `withListMeta` and the `ListMetaContributor` type. Deliberate under ADR-0010 (the barrel is an explicit named list, no `export *`); `tests/barrel.spec.ts`'s manifest was updated to match, which is what makes the addition a reviewed decision rather than a drive-by.
- **The generated OpenAPI list envelope** gains `required: ["items", "limit", "offset", "total"]` (deliberately not `meta`) and publishes `meta` as `additionalProperties: true` with a description, instead of a bare `{ type: "object" }`.

## Testing

`pnpm prettify`, `pnpm check`, `pnpm docs:links`, `pnpm format:check` and `pnpm docs:build` all run green. Real `pnpm check` tail:

```
> kavo@ depcruise
✔ no dependency violations found (304 modules, 1100 dependencies cruised)

> kavo@ lint
> oxlint packages examples tests .github/scripts

> kavo@ test
 Test Files  74 passed (74)
      Tests  1453 passed (1453)
```

Added:

- **`packages/core/tests/list-meta.spec.ts`** (new file) — driving the real registry override path, not the engine directly. Covers: the zero-config baseline (key absent, not `{}`); a handler's `meta` reaching the envelope; arbitrary JSON-serializable shapes (nested objects, arrays, `null`, booleans, numbers) landing verbatim; `meta` staying whole while `fields=` narrows the items; `withListMeta` adding keys without disturbing `entities`/`total`; the contributor receiving the real result and `KavoContext`; an async contributor; both merge-precedence directions; nested wraps; `meta` copied rather than aliased; `total ?? null` on the unwrapped path; and the wrong-shaped-handler error path parameterized over every branch of the message, asserting `KAVO_CONFIG_INVALID` and that the contributor never runs.
- **Omission cases** — an explicit `meta: {}` from a handler, and a whole wrap chain contributing nothing, both leave the key off; a `JSON.stringify` round trip pins `Object.keys` as exactly `["items", "limit", "offset", "total"]` with nothing contributed and `[..., "meta"]` with something.
- **`packages/core/tests/types/with-list-meta.test-d.ts`** — type-level acceptance that `FindManyResult["meta"]` is optional, that `ListResultDto["meta"]` is `ListMetaDto | undefined` while `total` narrows to `number | null`, that indexing `list.meta` without `?.` is a type error, that the contributor is entity-bound, and that `withListMeta(builtInHandlers(adapter)("findMany"), …)` drops into `operations.findMany.handler` **with no cast**.
- **`packages/frameworks/nest/tests/settings.e2e.spec.ts`** — wire-level: a contributed bag surviving the JSON round trip intact, and a plain controller's body having exactly `["items", "limit", "offset", "total"]`.
- **`packages/frameworks/nest/tests/binding.e2e.spec.ts`** — the published list schema marks `meta` as the one non-`required` envelope field, and documents it as an open bag (`additionalProperties: true`) rather than an object with no permitted keys.
- **`packages/protocols/mcp/tests/mcp-tools.spec.ts`** — a `findMany` tool call over a `withListMeta` service asserts `JSON.parse(result.content[0].text).meta`, pinning doc 16's "reaches the tool result unchanged".

Every behavior change was verified to fail against the old code before being kept: reverting `listMeta`'s omission turns 9 tests red across core and `@kavo/nest`, and reverting the schema change turns the two new Swagger tests red.

## Review notes

**Why `meta` is omitted rather than always `{}`.** `total` and `meta` pull in opposite directions on purpose. `total` reports `null` rather than disappearing when `pagination.count` is off, because every list answers "how many matched" and `null` *is* that answer. An empty `meta` answers nothing, and the zero-config list — the common case — is exactly what would pay for it on every response. Two details that fall out of that:

- **Omitted means omitted on both sides.** `mapResponse` spreads the key in conditionally rather than assigning `undefined`, because an `undefined`-valued key still appears in `Object.keys`/`in` for a programmatic caller even though `JSON.stringify` drops it. A REST client and a service caller must not disagree about whether the envelope has a `meta`.
- **Emptiness is judged after the merge**, so a contributor returning `{}` is indistinguishable from no contributor at all.

**`withListMeta` merge precedence — the contributor's keys win.** The wrapped handler's `meta` is the base and `compute`'s return value merges over it. Chosen deliberately for two reasons: it runs the same direction as every other precedence chain in Kavo (global → entity → operation → per-call — the more specific statement of intent wins), and it makes nested wraps layer predictably, so the outermost `withListMeta` owns any key it names. Keys the contributor doesn't name pass through untouched, so wrapping never drops what the inner handler set, and a contributor wanting the opposite has the inner bag in hand (`{ ...mine, ...result.meta }`). Both directions are pinned by tests and the choice is documented on the helper.

**Swagger: `meta` is the one envelope field with nothing to enumerate.** `items` is projected through the `list` DTO, so `schemaFromDto` reads real fields off it; `ListMetaDto` is an open `[key: string]: unknown` filled at request time by handler code Kavo never sees, so there is genuinely no static shape at decoration time. It is therefore published as `additionalProperties: true` with a prose description — a bare `{ type: "object" }` with no `properties` reads to most OpenAPI generators as an object permitting *no* keys, the opposite of the truth. Publishing a real per-entity `meta` schema would need a new declaration seam **and** a decision about whether it is documentation-only or an enforced projection (which would reverse doc 04's "meta is deliberately not a slot"). The GraphQL binding has the same open question for the same reason, so that decision is recorded in doc 10 §4 and doc 13's out-of-scope list rather than guessed at here.

**One judgement call worth a second opinion:** `withListMeta`'s parameter is typed `OperationHandler<Entity>` (exactly what `builtInHandlers(...)(id)` and `OperationConfig.handler` are) so the wrap composes with both without a cast — the ergonomics the helper exists for. That erases the inner output type, so a wrong-shaped inner handler is caught at run time with a `ConfigurationException` naming the operation instead of assembling a malformed envelope. The alternative — a precise `OperationHandler<Entity, null, FindManyResult<Entity>>` parameter — rejects `builtInHandlers(adapter)("findMany")` outright, which is the documented main path.

**GraphQL is deliberately out of scope** — and the issue's "Affected packages" is wrong to imply `meta` already flows to every transport. REST carries the raw envelope and MCP `JSON.stringify`s the whole result, so both get `meta` for free. But `packages/protocols/graphql/src/schema.ts` declares only `items`/`total`/`limit`/`offset` on the generated `<Entity>List` type, so **a GraphQL client cannot select `meta` today**. Exposing it needs a decision the other fields didn't (a `JSON` scalar vs. a per-entity metadata type), so rather than silently expanding scope, this PR records the gap in the GraphQL binding's own out-of-scope list and states the transport support explicitly in the adopter docs.

**Docs.** The adopter how-to lives in the Nest configuration reference next to `operations.<id>.handler`, and its example is deliberately explicit about where the adapter comes from: `@Kavo`'s config is evaluated at decoration time (ADR-0012), so `builtInHandlers` needs an adapter that exists by then — the module-scope `DataSource` the wiring guide already builds. The DI-factory (`forRootAsync`) caveat is called out rather than papered over. Every doc touching the word also disambiguates `ListResultDto.meta` (response bag) from `OperationConfig.meta` (`OperationMetadata`, ADR-0007).


mahdavipanah:
Pushed three commits addressing every non-blocking review finding (`61b67c2`, `b1b23a3`, `19fe555`).

**Correctness**

- **Malformed envelope could ship.** `isFindManyResult` checked only `entities`, so a wrapped handler returning `{ entities }` alone passed, `total` became `undefined`, `JSON.stringify` dropped the key, and the REST body shipped with no `total` — violating the normative envelope. No cast needed to hit it, since the parameter is deliberately output-erased. Fixed at both ends: the guard now also requires `total` to be a number or `null`, and `mapResponse` normalizes `total ?? null` so the unwrapped custom-handler path (same hole on `main`) is covered too. `describeValue` grew a branch so the message names *that* mistake rather than claiming `entities` is missing.
- **`meta` was handed out by reference.** The "matches `items`" justification was inexact — `items` is a fresh array of freshly serialized DTOs per request, whereas `meta` can be the very same object every time (a hand-written handler returning a module-scope constant is the documented alternative to the wrapper). `listMeta` now returns `{ ...handlerMeta }`, and the JSDoc states the accurate justification.

**Tests**

- `total ?? null` on the unwrapped path, plus both new guard rejections.
- `meta` is copied, not aliased — mutating one response leaves the handler's object and later responses pristine.
- `describeValue` was 100% unpinned (`return "x"` broke nothing; three of four branches never ran). The wrong-shape case is now parameterized over `null` / `[]` / `"nope"` / `{}` / the two `total` shapes, asserting each message fragment alongside `KAVO_CONFIG_INVALID`. Stubbing the function now fails six tests.
- `meta` with `total: null` — the two had never been observed co-occurring.
- MCP: a `findMany` tool call over a `withListMeta` service asserts `JSON.parse(result.content[0].text).meta`, pinning doc 16's "reaches the tool result unchanged".
- The type-level "Entity is inferred" assertion was vacuous (`result.entities.length` compiles against `readonly unknown[]`); it now maps the element type and asserts `readonly Author[]`.

**Docs**

- `using-the-api.md` claimed `total` is *omitted* when `pagination.count` is off. It is `null`, and the key is always present — pre-existing bug on the line adjacent to the new prose.
- The four landing-page list envelopes that still ended at `"total"` now carry `"meta": {}`, matching the canonical example.
- `ListMetaDto`'s doc comment still described the seam as hypothetical; it now names `FindManyResult.meta` and `withListMeta`.
- CLAUDE.md Conventions gained a line for higher-order wrappers (`with*`, never `wrap*`/`*Wrapper`) so a later synonym cannot appear beside `withListMeta`. The name itself is unchanged.

**Deliberately unchanged:** `withListMeta`'s parameter typing (the output-erased `OperationHandler<Entity>` is the correct call — the precise type rejects `builtInHandlers(adapter)("findMany")`), and GraphQL `meta` support (out of scope, correctly recorded in the docs).

Gate re-run green after restoring every experimental revert: `Test Files 74 passed (74)`, `Tests 1447 passed (1447)` (was 1436), `depcruise ✔ no dependency violations found (304 modules, 1100 dependencies cruised)`, plus `docs:links`, `format:check` and `docs:build`.

Each fix was verified to fail against the old behavior before being kept — reverting `total ?? null`, the shallow copy and the tightened guard turns exactly the five intended new tests red.

h16d:
Question on the envelope shape: `meta` is currently a required field (`readonly meta: ListMetaDto`), so every list response carries `"meta": {}` even when nothing contributes to it.

Worth considering omitting the `meta` key entirely from the JSON envelope when it's empty (and making the type `meta?: ListMetaDto`), to keep the wire payload free of noise for the common case where no handler/pagination strategy contributes anything.

Trade-off to weigh: that's a real reversal of this PR's stated design ("meta is a required envelope field... a handler that contributes nothing still yields an empty bag — never `undefined`"), and it'd touch the pinned "baseline: `meta: {}`" tests and the envelope-consistency argument (items/limit/offset/total/meta always present today). Not blocking — just flagging for discussion before merge.

h16d:
Follow-up on the Swagger side: `meta`'s OpenAPI schema currently renders as a bare `{ type: "object" }` with no `properties` (`listEnvelopeSchema` in `packages/frameworks/nest/src/swagger.ts`), unlike `item`/`list` which get real field schemas via `schemaFromDto`.

That's inherent to `ListMetaDto` being an open `[key: string]: unknown` bag with no static shape at decoration time — not a bug in this PR. But if per-entity meta schemas in Swagger are wanted later, it'd need a new seam (e.g. an entity-config-declared meta DTO/schema hint that `applySwaggerMetadata` reads, similar to `schemaForHint` for enum fields). Flagging as a possible follow-up issue rather than something this PR needs to solve.

mahdavipanah:
Worth flagging that `meta: {}` isn't new here — `mapResponse` already hardcoded it on `main`, and `@kavo/core@0.6.0` ships it. This PR only added it to the doc examples, which had drifted.

So omitting it would be a breaking change to a published envelope rather than a choice still open in this PR. My inclination is to keep it required: the envelope's shape then stays constant regardless of config (the same reason `total` is `null` rather than omitted), consumers write `res.meta.x` instead of `res.meta?.x`, and #118's `meta.nextCursor` would populate it for many configs anyway.

"Keep the wire lean" is a fair stance though — it just reads as envelope-wide (`total: null` too) rather than about this one field, so probably its own issue. Happy to discuss if you see it differently.

mahdavipanah:
Agreed, and agreed it's a follow-up — `swagger.ts` is untouched here.

One thing worth settling in that issue: unlike `item`/`list`, a declared meta schema wouldn't be enforced by anything. Those are trustworthy because the serializer projects through the DTO; `meta` is caller data that never does, and a contributor can legitimately vary its keys per request. So it's a fork rather than just a hint — a documentation-only declaration, or making `meta` a real DTO slot with projection (which would enforce it, but reverses the "meta is deliberately not a slot" call in doc 04).

The GraphQL binding has the same gap for the same reason (doc 13's out-of-scope list, same JSON-scalar-vs-typed choice), so it might be one issue covering the decision for both transports. No strong preference — open to whichever framing you prefer.

mahdavipanah:
@h16d — both of your points are now implemented (`2a469db`, `5a46b39`, `46ba382`). Reversing my earlier position on the first one: you were right that an empty bag on every list is noise the common case shouldn't pay for.

**1. `meta` is optional and omitted when empty.**

`ListResultDto.meta` is now `meta?: ListMetaDto`, and the engine leaves the key off the envelope entirely unless something contributed. A zero-config list body is exactly `{ items, limit, offset, total }`.

The envelope-consistency argument I made survives in a narrower form, and it's what decides `total` vs `meta`: `total` stays required and reports `null` when counting is off, because every list answers "how many matched" and `null` *is* that answer. An empty `meta` answers nothing. So the two fields diverge for a stated reason rather than by taste, and doc 07 §3.1 records it.

Two details worth flagging:

- **Omitted means omitted on both sides.** `mapResponse` spreads the key in conditionally rather than assigning `undefined` — an `undefined`-valued key still shows up in `Object.keys`/`in` for a service caller even though `JSON.stringify` drops it, and a REST client and a programmatic caller must not disagree about whether the envelope has a `meta`. Pinned both ways.
- **Emptiness is judged after the merge**, so a contributor returning `{}` is indistinguishable from no contributor. `withListMeta(h, () => ({}))` over a handler with no `meta` leaves the key off.

You were also right that this is a **breaking change to a published envelope**, not a choice still open inside this PR — `@kavo/core@0.6.0` ships `meta: {}`. It's marked `feat(core)!` with a `BREAKING CHANGE:` trailer and called out in the PR description: `result.meta.x` becomes `result.meta?.x`.

The pinned baselines flipped rather than being deleted: `engine.spec.ts` and the two e2e suites now assert absence, and there are new cases for the explicit-`{}` and empty-wrap-chain paths plus a `JSON.stringify` round trip asserting `Object.keys` exactly.

**2. Swagger.**

Both halves of the gap are addressed, and the second one turned out to be a real defect rather than just a missing nicety:

- `listEnvelopeSchema` now emits `required: ["items", "limit", "offset", "total"]` — deliberately not `meta`. Without it a generated client types `meta` as always-present, which is now wrong.
- `meta` is published as `{ type: "object", additionalProperties: true, description: … }` instead of a bare `{ type: "object" }`. That bare form has no `properties` *and* no `additionalProperties`, which most OpenAPI generators read as an object permitting **no** keys — the opposite of an open bag. `additionalProperties: true` says what it actually is. Two tests pin both.

What I did **not** do is the declared-per-entity-meta-schema seam. That's still the fork you and I landed on: a documentation-only declaration that nothing enforces (unlike `item`/`list`, which are trustworthy because the serializer projects through them), or making `meta` a real DTO slot with projection — which would enforce it but reverse doc 04's "meta is deliberately not a slot" and take away a contributor's ability to vary keys per request. That decision is recorded in doc 10 §4 and doc 13's out-of-scope list so both transports' gap points at the same open question, and I'm happy to open the joint issue for it — say the word and I'll file it against both bindings.

Gate re-run green on the pushed head: `Test Files 74 passed (74)`, `Tests 1453 passed (1453)`, `depcruise ✔ no dependency violations found (304 modules, 1100 dependencies cruised)`, plus `docs:links`, `format:check` and `docs:build`.

Each change was verified to fail against the old behavior before being kept: reverting `listMeta`'s omission turns 9 tests red across core and `@kavo/nest`, and reverting the schema change turns the two new Swagger tests red.

mahdavipanah:
Follow-up from a review pass on the pushed branch (`2d77944`) — one real defect in the omit-when-empty rule, caught by exactly the case #118 will write.

**`meta` keys whose value is `undefined` were counted as a contribution.** `listMeta` judged emptiness with `Object.keys(copy).length`, which counts a key holding `undefined` — precisely the key `JSON.stringify` erases. So a cursor contributor written the documented way, `withListMeta(h, () => ({ nextCursor: cursorOrUndefined }))`, produced `{ nextCursor: undefined }` on its last page: the envelope kept a `meta`, serialization emptied it, and the REST client received `"meta": {}` while the programmatic caller saw `{ nextCursor: undefined }`. That is the exact two-views divergence the rule exists to prevent, reintroduced through the other door.

The copy now skips `undefined` values, so an all-`undefined` bag omits `meta` entirely and a mixed bag drops only the erased keys — `{ exhausted: true, nextCursor: undefined }` ships as `{ exhausted: true }`. Two new cases pin both halves; each fails against the previous implementation.

Also fixed: the hand-written-handler snippet in the Nest configuration reference did not compile. `builtInHandlers(...)("findMany")` is `OperationHandler<Book>`, whose output type is `unknown` — the same erasure `withListMeta`'s runtime shape check exists for — so spreading the awaited result raised TS2698 for anyone copying the documented alternative to the wrapper. It now narrows explicitly, with a comment naming the erasure as the reason.

Gate green on the pushed head: `Test Files 80 passed (80)`, `Tests 1623 passed (1623)`, depcruise clean, plus `format:check`, `docs:links` and `docs:build`.

h16d (approved):

PR #127, opened 2026-08-04
feat(core): add computed/virtual fields to entity responses

### LINKED ISSUE #121: Add computed/virtual fields to entity responses
## Context

Kavo's default item/list projection (`DefaultSerializer`, `packages/core/src/serialization/default-serializer.ts`) is built entirely from ORM metadata — "getters/methods never appear (they are not columns)" (`docs/internals/architecture/04-dto-system.md` §3). There is no supported way to expose a field derived from other fields (e.g. `fullName` from `firstName`/`lastName`, or a formatted/aggregated value) without it being a real database column. This is common enough (Rails, Django REST Framework, and most CRUD frameworks call it a "computed"/"virtual" field) to deserve first-class support rather than the current happy-path workaround (registering a DTO class that names a property a TypeORM entity's own class getter happens to satisfy — which only works for TypeORM class instances, silently does nothing for `@kavo/prisma`/`@kavo/mongoose`'s plain objects, and is unreachable through `fields=` selection since the `selectable` allowlist only derives from real columns).

## Acceptance criteria

- [ ] New `EntityConfig.computed?: Record<string, ComputedFieldDescriptor<Entity>>` (global → entity → operation precedence doesn't apply here — like `dto`/`relations`, this carries functions, so it's entity-config-only, resolved once at `createCrud` bootstrap). `ComputedFieldDescriptor<Entity>` carries at minimum `resolve(entity: Entity, context: KavoContext<Entity>): unknown`.
- [ ] `DefaultSerializer.project()` evaluates a declared computed field by calling its `resolve` function instead of reading `key in source` — works uniformly for every ORM adapter (class instances and plain objects alike), so **no ORM adapter changes are needed**: computed fields are a pure response-mapping concern, resolved after the entity is already fetched.
- [ ] A computed field declared in `EntityConfig.computed` is included in the entity-derived default `item`/`list` projection automatically (same "declare what you want" posture as GraphQL's opt-in mutations) — no DTO class registration required to see it in a response, though an explicit item/list DTO class may still narrow it out like any other field.
- [ ] A computed field is added to the `selectable` allowlist by default (so `fields=fullName` works) unless the descriptor opts out; it is **never** eligible for `filterable`/`sortable` allowlists — bootstrap-time `ConfigurationException` if a computed field name appears in a configured `filterable`/`sortable` allowlist (there is no column to translate to `WHERE`/`ORDER BY`).
- [ ] Bootstrap-time `ConfigurationException` if a computed field name collides with a real column/relation name on the entity — naming the entity, the field, and the collision (matching the existing config-error quality bar: name the entity, the key path, the offending value).
- [ ] Computed fields are never writable: `DefaultDeserializer`'s writable projection is unaffected (already derives from `metadata.fields` only), and this stays true even if a computed field name also appears in a registered `create`/`update` DTO class.
- [ ] `resolve()` receives the same `KavoContext<Entity>` a custom operation handler gets (`principal`, `correlationId`, etc.), so a computed field can vary by caller — but it runs synchronously per served item; document that an async/DB-hitting resolver would defeat batching and is discouraged (N+1 risk), same posture the perf auditor already applies to includes.
- [ ] Tests cover: computed field present in default `findOne`/`findMany` responses; narrowed out by an explicit DTO that omits it; selectable via `fields=`; rejected (`ConfigurationException`) when declared as filterable/sortable; rejected when colliding with a real column name; absent from `create`/`update`/`patch` deserialization even when named in those DTOs; a computed field on a relation's target entity resolves correctly when that relation is included.
- [ ] `pnpm check` passes.
- [ ] `docs/internals/architecture/04-dto-system.md` gets a new section documenting computed fields (superseding the current "getters/methods never appear" line, which becomes wrong), and `docs/using-the-api.md` documents `EntityConfig.computed` for adopters.

## Affected packages

`@kavo/core`, docs — no ORM adapter changes (computed fields resolve after fetch, independent of the adapter), no `@kavo/nest` changes (Swagger already documents whatever DTO class shape is registered, and GraphQL/MCP resolvers already call the same bound service, so they pick up computed fields for free through the engine's serializer).

## Constraints

- DTO/serialization order stays normative (`docs/internals/architecture/04-dto-system.md` §5): DTO mapping first, then field selection — a computed field is subject to the same "selection narrows, never widens" rule as any other field.
- Security allowlist posture (`resolve-entity-config.ts`): anything outside `selectable`/`filterable`/`sortable` is a 400, never a silent drop — computed fields must slot into that same model, not bypass it.
- Core purity (ADR-0001, ADR-0005): the descriptor type and resolution logic live in `@kavo/core`; no ORM-specific typing leaks into `ComputedFieldDescriptor`.
- Naming (`CLAUDE.md`): no `I` prefix on the new interface; config key stays camelCase.

## Out of scope

- Automatic TypeScript type-level derivation of computed fields onto the entity-derived default response type — a caller wanting full static typing still registers an explicit item/list DTO class naming the field, as already required for any DTO narrowing today.
- Computing a field inside the database (a real generated/computed SQL column) — that's already supported today as an ordinary column via each ORM's own mechanism; this issue is specifically about fields with no backing column.
- Making computed fields filterable/sortable via in-memory post-fetch filtering — explicitly rejected above, not deferred.

mahdavipanah:
## What & why

Kavo's default `item`/`list` projection is built entirely from ORM metadata, so a field derived from other fields — `fullName`, a formatted total, a flag that depends on the caller — had no supported home. The happy-path workaround (register an `item` DTO naming a property the entity class happens to expose as a getter) only worked because `DefaultSerializer` did `if (key in source)`, which is an accident of TypeORM handing the engine class instances: `@kavo/prisma` and `@kavo/mongoose` return plain objects and the same config silently emitted nothing. Nor was the field ever reachable through `fields=`, since `selectable` derives from columns alone.

This adds `EntityConfig.computed` — a record of `ComputedFieldDescriptor<Entity>` keyed by the name each serializes as. `DefaultSerializer` **evaluates** a computed key by calling `resolve(entity, context)` instead of reading it off the row, which makes it uniform across every adapter. Declared fields join the entity-derived projection and the `selectable` allowlist automatically; they are never filterable, sortable, or writable, and those three limits are enforced rather than documented.

No ORM adapter changes were needed, and the reason is stronger than the issue argued: **no adapter consumes `query.fields` at all** (there is zero select pushdown — selection is "kept internally, stripped late"). Rows always arrive fully hydrated, so a computed field's source columns are present even under `fields=fullName`, and no dependency tracking is required.

Closes #121

## Public API impact

**Additive; nothing breaking for callers.**

- **New barrel exports** (ADR-0010's explicit list, plus `tests/barrel.spec.ts`'s manifest): `ComputedFieldDescriptor`, `ComputedFieldMap`.
- **Constructor signatures — the decision (correction #2).** `DefaultSerializer` and `DefaultDeserializer` each take a **new optional third parameter**, `computed: ComputedFieldMap<Entity> = {}`. Both are barrel exports, so this is deliberate: an optional trailing parameter keeps every existing `new DefaultSerializer(metadata, catalog)` call compiling and behaving identically.
  The alternative — reading `context.config.computed` per call — was rejected as *less* explicit, not more: the serializer already takes the root entity's `metadata` in its constructor rather than off the context, and a hand-constructed serializer should behave predictably regardless of what context it is handed. The relation-target path needed **no** signature change at all, since `projectionFor` already pulls the target's `ResolvedEntityConfig` off the `EntityCatalog` and that config now carries `computed`.
- **`ResolvedEntityConfig` gains a required `computed` member.** Kavo produces this type and the serializer/deserializer consume it; the only thing this breaks is code constructing one by hand. Noted in ADR-0019's consequences.
- **`EntityConfig`, `KavoInstance.createCrud`, `createCrud`, and `@Kavo` each gain a trailing type parameter** `Computed extends string = never`. It is inferred, defaults to the pre-existing behavior, and is invisible to every existing call site.

## Testing

`packages/core/tests/computed-fields.spec.ts` (25 cases) covers every criterion the issue lists: present in default `findOne`/`findMany`; evaluated off a plain object (not just a class instance); narrowed out by an item DTO that omits it *and* still evaluated when one names it; selectable through both the typed and wire fieldsets; `selectable: false` opting out into a 400 rather than a silent drop; never joining the filterable/sortable defaults; `KAVO_CONFIG_INVALID` for the filterable, sortable, column-collision, relation-collision and missing-`resolve` cases; stripped from create/update/patch bodies; and resolved on an included relation target through the catalog. `resolve` receiving the real `KavoContext` (principal, correlation id) is pinned too.

`packages/core/tests/types/computed-fields.test-d.ts` pins the inference, with `@ts-expect-error` on `filterable`/`sortable` and on a name no `computed` entry declares.

`packages/frameworks/nest/tests/binding.e2e.spec.ts` gains a wire-level block, which is the actual evidence for the "no `@kavo/nest` changes needed" claim rather than an assertion of it.

Verified locally:

```
pnpm check    →  74 test files passed, 1448 tests passed
                 (generate + build + typecheck + depcruise + lint + test, all green)
pnpm docs:links →  doc links OK — every tracked docs/**.md reference and sidebar link resolves
pnpm docs:build →  build complete
pnpm format:check → All matched files use Prettier code style!
```

## Review notes

**1. The deserializer subtlety (the issue got the rationale wrong).** The issue justified "computed fields are never writable" with "`DefaultDeserializer`'s writable projection is unaffected (already derives from `metadata.fields` only)". That reasoning is false: `deserialize` does `const allowed = dtoShapeKeys(dto) ?? this.writableProjection` — a registered `create`/`update` DTO **replaces** the projection entirely, so a DTO class declaring `fullName` for documentation would have passed the key straight through to the adapter as an unknown write. The requirement stands, but it took new code: the deserializer now holds the computed names and strips them whichever projection is in force. Two tests fail without that strip (verified by disabling it) — `"stays stripped when a registered create DTO names it"` and the direct `DefaultDeserializer` case.

**2. The type-level decision (correction #3).** An explicit `allowlists.selectable` is typed `readonly FieldPath<Entity>[]`, and a computed name is not a path on `Entity` — so without a change, configuring one meant a cast.

I threaded a trailing type parameter, `Computed extends string = never`, through `EntityConfig` → `createCrud` → `@Kavo`. It is inferred from the keys of `computed` (`Readonly<Record<Computed, ComputedFieldDescriptor<Entity>>>`), and `allowlists` sits in a `NoInfer<Computed>` position so listing a real column there cannot widen it and demand a matching `computed` entry.

It widens **only** `selectable`, to `FieldPath<Entity> | Computed`. `filterable`/`sortable` stay typed to real paths, which turns the never-filterable/never-sortable invariant into a compile error *before* it is a bootstrap error — the runtime `ConfigurationException` remains, for erased or cast configs (`@Kavo` erases to `EntityConfig<object>`) and for JS callers.

The tradeoff, stated plainly: a call site that pins all seven earlier generics explicitly gets `Computed = never` and is back to needing a cast for `selectable`. That is rare (the parameters exist to be inferred) and it is the price of not widening `QueryFieldSelector` to `string`, which would have thrown away the misspelling check that `usage-full.test-d.ts` pins today. The issue's "no type-level derivation" stays out of scope in the sense that matters: the *response* type does not grow the key, so a caller wanting it statically typed still registers an `item`/`list` DTO naming it — the same requirement as for any other narrowing.

**3. The ADR (correction #4).** [ADR-0019](docs/internals/adr/0019-computed-fields-are-serializer-evaluated.md) — *Computed fields are serializer-evaluated, and never filterable, sortable, or writable*. It records why the post-fetch placement is the only one available (nothing before response mapping sees both the hydrated entity and the context), and why the filter/sort limit is a permanent rejection rather than a deferral: in-memory predicates silently break pagination (`limit`/`offset` are applied by the database, before the predicate) and `total`. It is wired into the sidebar, into doc 04 §3 (whose "getters/methods never appear" line it supersedes, replaced by a new §7), into doc 05 §4, and into the code comments at each enforcement point.

**4. Left out deliberately.** Automatic type-level derivation of computed keys onto the entity-derived response type (out of scope per the issue, and it would mean synthesizing a response type from a config value, which no other DTO slot does); in-memory post-fetch filtering (explicitly rejected, not deferred); async resolvers (documented as an N+1 hazard, not offered).

**5. Note on a concurrent branch.** `feat/122-list-result-meta` also adds a barrel export. A trivial conflict in `packages/core/src/index.ts` (and `tests/barrel.spec.ts`'s manifest) at merge time is expected.

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -191,6 +191,10 @@ const config = defineConfig({
                 text: "0018 — Mongoose models are entity identities",
                 link: "/internals/adr/0018-mongoose-models-are-entity-identities",
               },
+              {
+                text: "0019 — Computed fields are serializer-evaluated",
+                link: "/internals/adr/0019-computed-fields-are-serializer-evaluated",
+              },
             ],
           },
         ],
--- a/docs/integrations/nest/configuration.md
+++ b/docs/integrations/nest/configuration.md
@@ -180,12 +180,20 @@ Response fields with no backing column, derived from an entity that has already 
 
 | Field        | Type                                        | What it does                                                                                         |
 | ------------ | ------------------------------------------- | ---------------------------------------------------------------------------------------------------- |
-| `resolve`    | `(entity, context: KavoContext) => unknown` | Derives the value. Called once per served item, **synchronously** — see the N+1 caveat below.        |
-| `selectable` | `boolean` (default `true`)                  | Whether `fields=` may name the field. `false` keeps it always-present and never individually chosen. |
+| `resolve`    | `(entity, context: KavoContext) => unknown` | Derives the value. Called once per served item, **synchronously** — see the caveats below.           |
+| `selectable` | `boolean` (default `true`)                  | Whether `fields=` may name the field. `false` makes naming it a 400 — read the note below carefully. |
 
 A declared computed field is in the default `item`/`list` projection with no DTO registration, and in the `selectable` allowlist by default. It is **never** filterable, sortable, or writable — naming one in `allowlists.filterable`/`sortable` is both a type error and a bootstrap `ConfigurationException`, and a value for one in a request body is stripped even if a registered `create`/`update` DTO declares the key.
 
-Keep `resolve` a pure function of the entity (plus `context.principal` where a field has to vary by caller). It runs per row, so a resolver that queries the database or calls out over the network reintroduces exactly the N+1 that batched includes exist to avoid.
+`resolve` returning `undefined` omits the key; `null` emits it — the same distinction a column draws.
+
+**What `selectable: false` does and does not mean.** It removes the name from the allowlist, so `?fields=auditNote` is a 400. It does **not** pin the field into every response: selection narrows the projection uniformly, so any request that sends `fields=` at all still drops it, and the client has no way to ask for it back. Read it as "not individually selectable", not "always present". An explicit `allowlists.selectable` list naming the field overrides the flag — an explicit list is always the deliberate answer.
+
+Keep `resolve` a pure function of the entity (plus `context.principal` where a field has to vary by caller). It runs per row, so a resolver that queries the database or calls out over the network reintroduces exactly the N+1 that batched includes exist to avoid. Declaring it `async` is a bootstrap error rather than a slow success: the serializer never awaits, so the promise would be emitted as-is and serialize to `{}`.
+
+On an **included relation target**, `resolve` receives the _root_ request's `KavoContext` — serving `GET /posts/1?include=author` hands an `Author` computed field a context whose `entityName`, `operation`, `config` and `query` describe Post. Only `principal`, `correlationId`, `transaction` and `state` mean what they say from a relation target.
+
+Pass the config inline to `@Kavo(...)` (or use `satisfies`) rather than annotating it `EntityConfig<Book>`: the computed-key type parameter is inferred at the call site, and an explicit `EntityConfig<Book>` annotation fixes it to `never` — which erases `computed`'s value types and leaves `resolve`'s parameter implicitly `any`.
 
 See [ADR-0019](/internals/adr/0019-computed-fields-are-serializer-evaluated) for the reasoning and [DTO system §7](/internals/architecture/04-dto-system) for how it interacts with DTO narrowing and field selection.
 
--- a/docs/integrations/nest/configuration.md
+++ b/docs/integrations/nest/configuration.md
@@ -163,7 +163,7 @@ What a request may filter, sort, and select on — including relation paths. Any
 | `sortable`   | same shape                                                    | Fields usable in `sort=`.       |
 | `selectable` | same shape                                                    | Fields usable in `fields=`.     |
 
-`{ exclude: [...] }` means "every own column except these", resolved against entity metadata at bootstrap. Omit a key entirely and it derives from the `query` DTO or entity metadata instead.
+`{ exclude: [...] }` means "every own column (plus, for `selectable`, every selectable computed field) except these", resolved at bootstrap against exactly the base set that key's plain default uses. Omit a key entirely and it derives from the `query` DTO or entity metadata instead.
 
 ### `computed`
 
@@ -172,7 +172,7 @@ Response fields with no backing column, derived from an entity that has already 
 ```ts
 @Kavo(Book, {
   computed: {
-    displayTitle: { resolve: (book) => `${book.title} (${book.year})` },
+    displayTitle: { resolve: (book) => (book.title === null ? null : `${book.title} (${book.year})`) },
     canEdit: { resolve: (book, context) => book.ownerId === (context.principal as User)?.id },
   },
 })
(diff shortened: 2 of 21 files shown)


mahdavipanah:
## Review round applied (`kavo-reviewer` + `kavo-test-auditor`)

Both auditors ran against the branch; everything blocking is fixed in `d91d680`. `pnpm check` → **74 files, 1461 tests passed**; `docs:links`, `docs:build`, `format:check` all green.

### Runtime fixes — three ways this could have been silently wrong

1. **`resolve` returning `undefined` produced an own key with value `undefined`** — invisible over HTTP (JSON drops it), visible to a programmatic caller, so behavior differed by transport. It now omits the key; `null` still emits, the same distinction the column branch draws.
2. **`__proto__` as a computed name silently vanished.** `resolved[name] = descriptor` sets the accumulator's prototype rather than adding a key, so the declaration disappeared with no error and `hasOwnProperty` in the serializer then read `false`. Now a bootstrap `ConfigurationException` — same class of hazard as 9a96a9c's bracket-segment fix.
3. **`async resolve` is now a bootstrap error.** ADR-0019 declared `resolve` synchronous, but nothing enforced it: the serializer emits the return value as-is, so an async resolver put a pending promise in the response — `{}` once serialized. `resolveComputedFields` now rejects `AsyncFunction`, turning a silently wrong body into a fail-fast. (A plain function that happens to return a promise still gets through; that is the limit of what is detectable there, and it is said so in the code.)

### Two claims that overstated what the code does

4. **`selectable: false` was documented as "always present".** It is not: it removes the name from the allowlist, but any request sending `fields=` at all still drops the field, and the client cannot ask for it back. Corrected in the JSDoc, the Nest configuration reference, and a new ADR-0019 consequence; a test pins the omitting-fieldset case in both directions.
5. **A relation target's resolver receives the *root* request's context.** `ComputedFieldDescriptor<Entity>.resolve` types it `KavoContext<Entity>`, but `GET /posts/1?include=author` hands an `Author` field a context whose `entityName`, `operation`, `config` and `query` are Post's — with full type-checker blessing. Documented on `resolve`, in the configuration reference, and in ADR-0019, which also records why a per-target context was rejected: it would have to invent an `operation` no caller issued and a `query` never normalized against that entity. A test now pins the semantics.

### Docs sync

`docs/internals/architecture/08-configuration.md` was stale in four places (§2 structural-key list, §3 bootstrap-contract enumeration, §4 validation, §6 debug dump) — §2 literally disagreed with the code comment this PR had already updated. Fixed.

### Tests

- The relation-isolation test was **vacuous**: it asserted the absence of a name no entity in the fixture declared, and passed identically if the serializer used the root's computed table for the relation node. Rewritten — `label` is now declared on Post *and* Author with different resolvers, so each side using its own table is an assertion rather than an absence.
- Config errors now assert `messageParams` (entity + key path), matching `config-validation.spec.ts`'s bar, so the key path can't drift.
- New: to-many include (the `list`-slot lookup in `projectionFor`, a distinct line); target's own item DTO naming *and* omitting the computed field; `allowlists.selectable: { exclude }` in both directions; explicit list beating `selectable: false`; a throwing resolver's problem details (`KAVO_PERSISTENCE_FAILED`, no internal message leaked); wire-level `undefined` omission; and `@Kavo`'s new `Computed` parameter in the Nest type tests, with `@ts-expect-error` on filterable/sortable.

### Deliberately not changed

- **`ResolvedEntityConfig.computed` stays required.** Making it optional would let the serializer silently see `undefined`. Flagged in the PR body and in ADR-0019's consequences; there is no CHANGELOG in this repo to add a line to.
- **The `EntityConfig<Entity>` annotation wrinkle** — annotating a config `EntityConfig<Book>` fixes `Computed` to `never` and erases `computed`'s value types. Inference at the `createCrud`/`@Kavo` call site is the dominant path and works; the configuration reference now says to pass the config inline (or `satisfies`) rather than annotating it.
- **Doc 03's contract inventory** already omits `QueryAllowlists`/`QueryFieldSelector`, so it isn't maintained as exhaustive — left alone rather than half-updated.

mahdavipanah:
Review findings addressed — 4 commits pushed (`2288a7f`, `9a86d54`, `6ca27eb`, `9580a21`). Gate green: `pnpm check` 74 files / 1469 tests, plus `docs:links`, `docs:build` and `format:check`.

### Blocking

**B1 — resolve-beats-row precedence was unpinned.** Two cases added: a seeded row that *carries* the computed key (`{ id: 1, name: "Ada", shout: "FROM ROW" }`) and a subclass exposing it as a class getter — ADR-0019's own motivating example, since that getter workaround is exactly what computed fields replace. Verified: swapping the two branches in `DefaultSerializer.project` fails **exactly** these two (`FROM ROW` / `FROM GETTER`) and nothing else.

**B2 — never-writable strip pinned for `create` only.** Superseded by finding 6 (below). A write DTO naming a computed key is now a bootstrap error, so the runtime strip is unreachable through `createCrud` and the DTO cases became bootstrap-error assertions across all three slots (`create`, `update`, `patch`), plus a positive case so the check isn't just "any DTO". The strip stays as the inner layer for a directly-constructed `DefaultDeserializer`, and its direct-unit test is what fails when the line is deleted — verified: removing `if (this.computedNames.has(key)) continue;` fails exactly that one test. `binding.e2e.spec.ts` gained the bind-time rejection instead of a restatement.

### Non-blocking

3. `{ exclude }` prose fixed in `05-query-grammar.md` and the Nest config reference — "every own column (plus, for `selectable`, every selectable computed field) except the ones named".
4. `resolveFieldSelector`'s `ownColumns` → `base`, docstring reworded to "the allowlist key's base set".
5. `__proto__` prototype check added — the object-literal spelling invokes the prototype setter and never reaches `Object.keys`, so it registered nothing and threw nothing. Test added; verified it fails without the check.
6. **A write DTO naming a computed key is now a bootstrap `ConfigurationException`** at `dto.create`/`dto.update`/`dto.patch`, consistent with every other computed misdeclaration, and noting the OpenAPI consequence (`@ApiBody` is built from the DTO's runtime shape). Both layers kept.
7. Computed fields absent from the generated OpenAPI response schema — documented in ADR-0019's Consequences and the Nest `computed` reference: register an `item`/`list` DTO, the same escape hatch as for the static type.
8. **`resolve` must be total, not merely pure** — one throwing row kills the whole list response. ADR-0019 §1, the config reference, `using-the-api`, doc 04 §7 and the JSDoc all say so; every example resolver in docs and tests is now defensive. The serializer still does not swallow the throw.
9. `resolve` receives the fully hydrated row — documented as deliberate, and as making the resolver part of the exposure decision.
10. Throwing resolver → `KAVO_PERSISTENCE_FAILED` recorded in ADR-0019, with the note that response mapping runs *after* a committed write on `createOne`, so a retry duplicates.
11. `Computed = never` note now covers explicit type arguments (`@Kavo<Todo, CreateTodoDto>(...)`) as well as the `EntityConfig<Book>` annotation.
12. Vacuous e2e test deleted (it asserted a property of `JSON.stringify`, not of Kavo).
13. `hasOwnProperty`-not-`in` guard claimed by a test (an `item` DTO naming `toString`) — verified it fails when mutated to `in`.
14. ADR-0019's negative type claim added to `computed-fields.test-d.ts`.
15. Perf: `relations` and the fieldset are `Set`s built once (constructor / per response / per relation node) rather than rescanned per key per row; `narrowToDto`'s allocation hoisted out of the per-row loop by giving `serializeList` its own hoisted body. Deselected-resolver-never-called test added — verified it fails when the selection check is moved after the computed branch.

ADR numbering unchanged: this PR keeps **0019**.

h16d (approved):

PR #128, opened 2026-08-04
feat(core,nest): ETag response caching and If-Match preconditions

### LINKED ISSUE #120: Add ETag response caching and If-Match precondition checks
## Context

Kavo has no HTTP caching or optimistic-concurrency support today: `KavoRequest`/`KavoResponse` (`packages/core/src/context/`) carry no header-like preconditions, and `@kavo/nest`'s generated routes never set response headers or read conditional-request headers. Adding ETags lets clients avoid re-downloading unchanged resources (`If-None-Match` → `304 Not Modified`) and lets writers detect they're overwriting a version they never saw (`If-Match` mismatch → rejected, not silently applied). Per user decision, the ETag is a **content hash of the serialized item** (not a config-declared version column) — this keeps the whole feature inside `@kavo/core` + `@kavo/nest`, with zero ORM adapter changes, and it is **enabled by default** via a new global setting.

## Acceptance criteria

- [ ] New `KavoSettings.caching: { etag: boolean }` (default `{ etag: true }`) — global → entity → operation → per-call precedence, same as every other settings key.
- [ ] Every single-item response (`findOne`, `createOne`, `updateOne`, `patchOne`, `restoreOne`) gets a deterministic hash-based ETag computed from the serialized item DTO, exposed on `KavoResponse` for the transport layer to turn into an HTTP `ETag` header. Collection (`findMany`) responses are out of scope (see below).
- [ ] `KavoRequest` gains a transport-agnostic way to carry incoming conditional-request tokens (e.g. `preconditions: { ifMatch?: readonly string[], ifNoneMatch?: readonly string[] }`), populated by `@kavo/nest` from `If-Match`/`If-None-Match` headers; programmatic callers may also pass it directly.
- [ ] `findOne`: when `If-None-Match` matches the current computed ETag, the engine signals "not modified" on `KavoResponse` and `@kavo/nest` returns `304 Not Modified` with no body, instead of re-serializing and re-sending the item.
- [ ] `updateOne`/`patchOne` (and `deleteOne`, per RFC 9110 — `If-Match` is a standard precondition on any write, not just `PUT`/`PATCH`): when `If-Match` is present and does not match the entity's current computed ETag, the write is rejected before being applied — new `PreconditionFailedException` / `KAVO_PRECONDITION_FAILED` error code, mapped to HTTP `412 Precondition Failed` in the error catalog (`packages/core/src/errors/error-catalog.ts`), following the same pattern as `ConflictException`/`KAVO_CONFLICT`. A missing entity still raises `NotFoundException` (404) — precondition evaluation happens only once the entity is confirmed to exist.
- [ ] Documented, explicit limitation: since the ETag is a content hash rather than a DB-enforced version column, the mismatch check is application-level check-then-write, not an atomic compare-and-swap — a true race window exists between the check and the adapter's write. State this plainly in docs rather than implying DB-level atomicity; closing that gap (e.g. a config-declared version column with adapter-level conditional writes) is explicitly out of scope here.
- [ ] `@kavo/nest`'s generated routes read `If-Match`/`If-None-Match` request headers and set the `ETag` response header / `304` status, without requiring hand-written controller methods for every entity (manual-method-wins still applies as it does everywhere else).
- [ ] Tests cover: ETag present on every single-item response; `If-None-Match` hit → `304`, no body; `If-None-Match` miss → normal `200` + fresh ETag; `If-Match` hit on write → write applied, new ETag returned; `If-Match` miss on write → `412`, write not applied, entity unchanged; `If-Match` on a missing entity → `404`, not `412`; `caching.etag: false` at any config scope disables ETag generation and precondition checks entirely.
- [ ] `pnpm check` passes.
- [ ] Docs: `docs/using-the-api.md` (or a new page) documents the `caching.etag` setting, the conditional-request headers supported, the 304/412 behavior, and the check-then-write race-window limitation above.

## Affected packages

`@kavo/core`, `@kavo/nest`, docs — no ORM adapter changes (`@kavo/typeorm`, `@kavo/prisma`, `@kavo/mongoose`, `@kavo/mikroorm` are all unaffected, since the ETag is computed from the serialized response, not read from the database).

## Constraints

- Config keys stay camelCase, booleans phrased positively (`CLAUDE.md`) — `caching.etag`, not `caching.disableEtag`.
- Core purity (ADR-0001, ADR-0005): `KavoRequest`/`KavoResponse`/the new exception all live in `@kavo/core`; hashing must use only Node/runtime built-ins Kavo already depends on (core ships zero runtime dependencies) or be dependency-free.
- Problem-details error shape (ADR-0009): the new `KAVO_PRECONDITION_FAILED` code follows the existing catalog format (`status`, `title`, `message` template) next to `KAVO_CONFLICT`.
- Route generation happens at decoration time (ADR-0012) — header/status handling must fit the existing generated-route shape, not require per-entity hand-written controllers.

## Out of scope

- Collection-level (`findMany`) ETags/caching.
- A config-declared version/`updatedAt` column and true atomic compare-and-swap writes at the adapter level (would touch every ORM adapter — tracked as a possible follow-up if the content-hash approach's race window proves insufficient).
- `Cache-Control`/`Last-Modified`/other HTTP caching headers beyond `ETag`.
- GraphQL/MCP/gRPC protocol bindings picking up equivalent caching semantics — REST (`@kavo/nest`) only for this issue.

mahdavipanah:
## What

ETag response caching and `If-Match` preconditions, behind a new `KavoSettings.caching: { etag: boolean }` that defaults to `true`.

- Every single-item response (`createOne`, `findOne`, `updateOne`, `patchOne`, `restoreOne`) carries a strong `ETag` — a SHA-256 of the **canonicalized** serialized representation. Collection responses carry none.
- `If-None-Match` on a read → `304 Not Modified`, no body, same `ETag`.
- `If-Match` on `PUT` / `PATCH` / `DELETE` → the write is refused with `412 Precondition Failed` when no supplied token matches the target's current tag, and **nothing is written**. A missing target is `404`, not `412`.
- `caching.etag: false` at any scope (global / entity / operation / per-call) turns off tag generation *and* precondition evaluation together.

## Why

Kavo had no HTTP caching and no optimistic concurrency: clients re-downloaded unchanged resources, and a writer had no way to detect it was overwriting a version it never saw. Per the issue's decision the tag is a content hash rather than a config-declared version column, which keeps the whole feature inside `@kavo/core` + `@kavo/nest` — **zero ORM adapter changes**.

Closes #120

## Public API impact

**Breaking (type-level, for code that constructs these):**

- **`KavoResponse` gained `etag: string | null` and `notModified: boolean`**, both required. Reading a `KavoResponse` is unaffected; *constructing* one (a fake engine in a test, a custom transport) must now supply them. One in-repo site needed updating (`packages/core/tests/kavo-service.spec.ts`).
- **`KavoEngineDependencies` gained a required `reader: EntityReader<Entity>`.** Anything constructing a `KavoEngine` by hand must supply it. In-repo that is only `createKavo`, which already holds the adapter.

**Additive, non-breaking:**

- `KavoSettings.caching: CachingSettings` (`{ etag: boolean }`, default `true`) — a new required key on the *resolved* settings type; every input scope is `DeepPartial`, so no caller has to set it.
- `KavoRequest.preconditions?: RequestPreconditions | null` — optional, so every existing caller keeps compiling.
- `KavoCallOptions.preconditions?: RequestPreconditions` — how the typed service surface reaches the above without every method growing a parameter.
- New error code `KAVO_PRECONDITION_FAILED` (412) and `PreconditionFailedException`. Codes are API surface once shipped; this one is new, so nothing is renamed.
- Core barrel: `+ PreconditionFailedException`, `+ PreconditionUnsupportedException`, `+ type CachingSettings`, `+ type RequestPreconditions`. `computeEtag` stays internal — nothing in the workspace imports it, and exporting it would make SHA-256-of-canonicalized-JSON a public promise this ADR's own consequences say may be superseded (ADR-0010).
- `@kavo/nest` barrel: `+ KavoResponseInterceptor`, `+ ConditionalRequest`, `+ parseEntityTags`.

**Behavioral change worth flagging:** generated route handlers now go through `service.engine.execute(...)` and return the envelope, which a method-scoped interceptor unwraps — rather than delegating to the typed `DefaultKavoService` methods. It is the same pipeline (those methods *are* `execute` plus that unwrap), the wire output is unchanged. `@Override`'d and hand-written methods keep returning their own value — `KavoResponseInterceptor` is applied to them too but acts only on an engine envelope, so an override that returns `engine.execute(...)` keeps the `ETag` instead of silently losing it. CLAUDE.md and doc 10 are updated, since the old sentence is no longer true.

Also: `@Override`'d methods now receive one extra trailing parameter (the preconditions), because both paths deliberately carry identical route metadata. Declaring fewer parameters than the layout offers is already how an override opts out, so nothing existing breaks.

## Testing

New:

- `packages/core/tests/etag.spec.ts` (20 tests) — canonicalization properties (key order, array order, `null` vs. absent, `undefined`, dates, non-finite numbers, `toJSON`), tag shape/stability, and RFC 9110 strong vs. weak comparison including `*` and multi-candidate lists.
- `packages/core/tests/caching.spec.ts` (42 tests) — tags on every single-item response including `restoreOne`; none on collections or void results; per-representation tagging for both `fields=` and `include=`; `If-None-Match` hit/miss/wildcard and ignored-on-writes; `If-Match` hit applies, miss → `PreconditionFailedException` + `KAVO_PRECONDITION_FAILED`/412 with the row **asserted unchanged** in every arm, stale `DELETE`/`restoreOne`/`purgeOne` blocked, fresh ones allowed, missing target → `NotFoundException`/404, `DELETE` on a soft-deleted row keeping its own 409 header or no header, a tag from a `fields=`-narrowed read 412ing, a present-but-empty `If-Match` 412ing, weak tag rejected, wildcard accepted **with zero pre-reads**, current tag named in `messageParams`, the `KavoCallOptions` route; `If-Match` refused with `KAVO_PRECONDITION_UNSUPPORTED` on `createOne`, on a custom operation, and when `findOne` is disabled; and `caching.etag: false` at global/entity/operation/per-call scope, both halves each.
- `packages/frameworks/nest/tests/caching.e2e.spec.ts` (31 tests) — the same matrix over real HTTP through a real Nest app, plus header parsing, empty-header refusal, the two `@Override` arms (guard dropped vs. guard forwarded), and the envelope never leaking onto the wire. State assertions read the row back rather than counting rows, since `Todo` is soft-deletable and a row count passes on the wrong behavior.

The e2e suite **turns Express's own weak ETag off**, so any `ETag` or `304` it observes can only be Kavo's — Express both tags responses and 304s a fresh `If-None-Match` on its own, and a suite that left it on would have passed without Kavo doing anything. One test leaves it on, to pin that Kavo's tag wins.

Updated: `errors.spec.ts` and `barrel.spec.ts` manifests, `config-validation.spec.ts`, `kavo-service.spec.ts`, and `binding.e2e.spec.ts` (the generated OpenAPI now documents `ETag`, `If-Match`/`If-None-Match`, `412` and `304`).

**Verified gate** (`pnpm check`, run on this branch at 66d902e):

```
> depcruise packages examples --config .dependency-cruiser.cjs
✔ no dependency violations found (307 modules, 1116 dependencies cruised)

> oxlint packages examples tests .github/scripts

> vitest run
 Test Files  76 passed (76)
      Tests  1518 passed (1518)
```

`pnpm docs:links` and `pnpm docs:build` pass too.

`pnpm docs:links` → `doc links OK`. `pnpm docs:build` → `build complete`. `pnpm format:check` → `All matched files use Prettier code style!`.

## Review notes

The issue understated four things; here is what I decided and why.

**1. `node:crypto` is banned in core.** `.dependency-cruiser.cjs`'s `core-imports-nothing` forbids any import resolving outside `packages/core`, so the issue's "runtime built-ins Kavo already depends on" would have failed the gate. I used the in-repo precedent instead — a typed `globalThis` accessor, the way `default-kavo-context.ts` reaches `crypto.randomUUID` (ADR-0005) — for both `crypto.subtle.digest` and `TextEncoder`.

I chose **async SHA-256 over a sync FNV-1a**: for `If-Match`, a hash collision is a silently lost update, which is the one failure this feature exists to prevent, and 32-bit FNV's birthday bound is not something I want load-bearing there. The cost is that `mapResponse` is now `async`, which is free — `execute` already was.

**2. Canonicalization is explicit, not left unstated.** `computeEtag` sorts object keys before hashing, so a DTO field reorder is neither a spurious cache miss nor a spurious 412. Otherwise it follows `JSON.stringify`'s conventions — `toJSON` honored, `undefined` members dropped, dates as ISO strings — because the thing being hashed is what is about to be JSON-encoded, and a value type whose `toJSON` distinguishes states its enumerable keys do not would otherwise be an ETag collision.

**3. `KavoResponse` exposure alone does not reach the route — I took the interceptor route.** The typed service surface discards the envelope (`findOne(...): Promise<ItemDto>`), so exposing `etag` there was not enough. I switched generated handlers to `engine.execute` and added a method-scoped `KavoResponseInterceptor` (applied as an *instance*, so it needs no DI registration) that sets the header, converts `notModified` to a 304, and unwraps. That keeps **one** computation site in core and one unwrap site in nest, rather than hashing twice and comparing twice.

Setting the status from an interceptor works because Nest applies the route's static `@HttpCode` *before* interceptors run and passes `undefined` to `reply`, so it is not re-applied — I read `router-execution-context.js` to confirm this rather than assume it. `header`/`status` are duck-typed for Express/Fastify, per `kavo-exception.filter.ts`'s precedent.

Collapsing `makeHandler` to one arm was a bonus: the fixed parameter layout is now read from exactly the one place that writes it, and a custom read operation no longer has its query handed to it in the body slot.

**4. The `If-Match` pre-read had no home — that is ADR-0020.** I injected `reader: EntityReader` into `KavoEngineDependencies` and evaluate the precondition engine-side, since handlers have the adapter but not the serializer the hash needs. `deleteOne` gets the pre-read too.

`restoreOne`/`purgeOne` are **included**: the pre-read asks for `withDeleted` on a soft-deletable entity, which widens the filter and never touches the projection, so one read serves live and soft-deleted rows alike and produces the tag `GET /books/1?withDeleted=true` already served. Every operation the engine *cannot* evaluate — `createOne`, any custom operation, `caching.etag` off, `findOne` disabled — is **refused** with 412 `KAVO_PRECONDITION_UNSUPPORTED` rather than silently performed. Reads are the sole exception and ignore `If-Match`, since a safe method cannot lose an update.

**5. Representation-dependence: I chose per-representation, and documented the consequence.** RFC 9110 ETags identify a *representation*, so `GET /books/1?fields=title` correctly gets a different tag from `GET /books/1`. Preconditions are evaluated against the entity's **canonical read representation** — what `findOne` with no query params returns — which is where a client's tag comes from in practice. The consequence, stated in both the ADR and `using-the-api.md`: an ETag taken from a narrowed read will 412.

The alternative (always hash the full item) would make `If-Match` work from any read, but would give the same strong tag to two different representations, which a shared cache can get wrong. I preferred spec-conformance plus an explicit doc line.

**6. The race window is stated plainly.** This is application-level check-then-write, not compare-and-swap. Both the ADR ("it narrows the window a naive last-write-wins API leaves wide open; it does not close it") and the adopter page say so outright, and neither implies DB-level atomicity. A version column with adapter-level conditional writes stays out of scope.

## Conflicts

Expect a conflict with #122 in `mapResponse` — it is now `async` and takes a `preconditions` parameter, and both list and item branches gained `etag`/`notModified`. Nothing here touches #121's serializer or config-resolution area.


mahdavipanah:
Review findings addressed — pushed as five commits (`06a55eb…66d902e`).

### Blocking

**B1 — ADR renumbered 0019 → 0020.** File, heading, VitePress sidebar entry, and every citation in code and docs. `pnpm docs:links` and `pnpm docs:build` pass.

**B2 — `If-Match` no longer fails open.** Both suggested fixes, applied where each belongs:

- `restoreOne`/`purgeOne` are now evaluated. The pre-read asks for `withDeleted` on a soft-deletable entity — which widens the *filter* and never touches the projection — so one read serves live and soft-deleted rows alike, and the tag it produces is byte-identical to what `GET /books/1?withDeleted=true` already served. `PRECONDITION_TARGETS` is now the five standard single-row writes.
- Everything else that changes state is **refused**, not returned from: a new `PreconditionUnsupportedException` (412 `KAVO_PRECONDITION_UNSUPPORTED`) covers untargeted operations (`createOne`, every custom operation), `caching.etag: false` for the operation in force, and — see finding 8 — `findOne` not being enabled.
- **Reads are the one deliberate exception**: `If-Match` on a `GET` is ignored, not refused. A safe method cannot lose an update, so honoring it could not prevent anything, and refusing would break a legal-if-rare request for no safety gain. That is now stated in ADR-0020 §4, in `docs/using-the-api.md`, and pinned by a test — a decision rather than a leftover.
- On `caching.etag: false`: refusing, deliberately. Ignoring answers `2xx` for a guard nothing applied, and the per-operation scope makes the asymmetry easy to reach by accident. Documented in `configuration.md#caching` and ADR-0020 §6; the three existing "ignores an If-Match" tests now assert the refusal.

**B3 — replaced methods.** Semantics chosen and implemented: `KavoResponseInterceptor` now covers `@Override`'d methods too (it acts only on an engine envelope, so an override returning its own value is untouched), which means an override that returns `engine.execute(...)` keeps the `ETag` instead of silently losing it. Enforcement itself **cannot** follow — it lives in the engine, and a method that does not call the engine cannot have it applied on its behalf. So that half is documented loudly (ADR-0020 §7, doc 10, `@Kavo`'s own jsdoc, `configuration.md`) and pinned by two e2e tests: a bypassing `@Override() updateOne` (stale `If-Match` → 200, row overwritten, no `ETag`) and a forwarding one (stale → 412 + row unchanged; fresh → 200 + `ETag`). The example's `AddressController` now forwards its `preconditions` rather than demonstrating the bypass.

**B4 — `AlreadyDeletedException` preserved.** A `null` pre-read no longer raises 404 from the engine; it falls through to the handler, which raises the operation's own error. `DELETE` on a soft-deleted row is 409 `KAVO_ALREADY_DELETED` with or without the header. Both arms tested, in core and over HTTP.

**B5 — the 412 test now proves the write was blocked.** Re-`GET`s the row and `toMatchObject`s its fields, plus `adapter.rows.filter(r => r.deletedAt === null)`.

**B6 — coverage.** Added: stale `restoreOne`/`purgeOne` (and fresh, via a `withDeleted` tag); the per-call scope's `If-Match` half (the loop now carries per-call request options as a fourth entry); a `fields=`-narrowed tag fed to `patchOne` → 412 + row unchanged, so `canonicalEtag` cannot start honoring `request.query` unnoticed; an `include=`-narrowed representation getting its own tag; and state assertions on the weak-tag and bare-200 tests.

### Non-blocking

7. `parseEntityTags` now returns `[]` for a present-but-empty header and `undefined` only for an absent one — `If-Match:`, `If-Match: " "`, `If-Match: ,,,` all 412 with the row unchanged. The test at `:295` asserts the new behavior.
8. The pre-read is skipped and the request refused when `findOne` is not an enabled registry entry, so no hash of an unserved representation reaches the 412 `detail`. A test asserts no 64-hex-char string appears in the message.
9. `canonicalize` honors `toJSON`, applied once per node. Tested with a `Decimal`-shaped value whose wire form and enumerable keys disagree in both directions.
10. `computeEtag` dropped from the barrel. `RequestPreconditions` (type-only now), `CachingSettings` and `PreconditionFailedException` stay; `PreconditionUnsupportedException` is added.
11. `KavoCallOptions.preconditions` documents that only `ifMatch` does anything through the typed surface — `ifNoneMatch` is engine-only.
12. OpenAPI documents the `ETag` response header, `If-None-Match` + `304` on single-item reads, and `If-Match` + `412` on single-row writes, gated on as much of `caching.etag` as decoration time can see. Four new assertions in `binding.e2e.spec.ts`.
13. The interceptor-ordering hazard is documented in `configuration.md`, doc 10, the interceptor's own jsdoc, and ADR-0020's consequences.
14. `If-Match: *` short-circuits before the pre-read, and the `TextEncoder` is hoisted to module scope. A counting adapter pins zero pre-reads for `*` and exactly one for a real tag.
15. PR body corrected: 20 / 42 / 31, not 24 / 28 / 24.

### Gate

```
> depcruise packages examples --config .dependency-cruiser.cjs
✔ no dependency violations found (307 modules, 1116 dependencies cruised)

> oxlint packages examples tests .github/scripts

> vitest run
 Test Files  76 passed (76)
      Tests  1518 passed (1518)
```

`pnpm prettify`, `pnpm format:check`, `pnpm docs:links` and `pnpm docs:build` all clean.

### Not changed

- **`caching.etag` stays on by default** and no `findMany` ETags — both out of scope per the issue.
- **The 412 for an unevaluable precondition is a second code, not a reused one.** `KAVO_PRECONDITION_FAILED` says "re-read and retry"; retrying `KAVO_PRECONDITION_UNSUPPORTED` will never succeed, so the two need different codes even though both are 412. The catalog's existing comment about free-form message params is why the reason arrives as a closed set of phrases written at the single throw site, not as caller text.

h16d (approved):

PR #129, opened 2026-08-04
feat(core,typeorm,prisma,mongoose,mikroorm): cursor (keyset) pagination

### LINKED ISSUE #118: Add cursor-based pagination strategy

mahdavipanah:
> **Stacked on #126** (`feat/122-list-result-meta`) — **merge that first.** This PR targets `feat/122-list-result-meta`, not `main`, because it builds on #122's `ListResultDto.meta` seam. Once #126 lands, this retargets to `main` cleanly.

Closes #118

## What

Adds a third built-in pagination strategy, `pagination.strategy: "cursor"` — keyset paging over an opaque `?cursor=` token, with the next page's token returned as `meta.nextCursor` (`null` on the last page). Supported on all four ORM adapters.

## Why

Offset paging degrades on large tables (the database still walks the skipped rows) and is unstable under concurrent writes — rows shift between pages, so a client can miss or repeat one. Keyset paging is the standard fix and drops naturally into the existing `PaginationStrategy` seam, which was already typed open.

## Public API impact

**One consumer-side break**, at v0.6.0 lockstep. `Pagination` is now a **union**:

```ts
type Pagination<Entity = unknown> = OffsetPagination | CursorPagination<Entity>;
interface OffsetPagination { limit: number; offset: number }              // unchanged
interface CursorPagination<Entity = unknown> {
  limit: number; cursor: string | null; keyset: FilterExpression<Entity> | null;
}
```

- **Producers are unaffected.** The discriminant is the *presence of `cursor`*, not a `kind` tag, so `OffsetPagination` stays structurally identical to the old shape — every existing strategy, including third-party ones, is still assignable with no edit.
- **Consumers that read `.offset` must narrow first**, with the new `isCursorPagination` guard. That hits anything implementing `EntityReader`/`RepositoryAdapter`. It is unavoidable given the issue's own requirement that "an adapter never receives a meaningless offset" — only the type can enforce that — and it's exactly what ADR-0021 exists to justify.

New barrel exports: `OffsetPagination`, `CursorPagination`, `CursorPaginationStrategy`, `isCursorPagination`, `encodeCursor`, `cursorValuesOf`, `readFilter`. (`decodeCursor` and `keysetExpression` are deliberately *not* exported — they are `QueryNormalizer` internals whose contracts only hold once the effective sort has been validated.) `QueryContext` gains `cursor?: string | null`; `FindManyResult` gains `hasMore?: boolean`; `PaginationStrategyName` gains `"cursor"`. `offset` remains the default strategy — nothing changes for an entity that doesn't opt in.

## Review notes — the five corrections, and how they were resolved

**1. Opaque, not tamper-resistant.** Confirmed unachievable as the issue worded it: `core-imports-nothing` blocks `node:crypto`, `SubtleCrypto` is async against a sync `normalize`, and `KavoSettings` has no key concept. Implemented as **base64url-encoded JSON, strictly shape-validated on decode** — arity must match the effective sort, and every value must match the field's declared `FieldKind` (enum members checked against the declared set, dates must parse, `null` refused). Failures are the existing `KAVO_QUERY_INVALID_VALUE` issue on `cursor`.

Nothing in the code or docs claims tamper resistance. Both state the real argument: a cursor payload is a tuple of comparison values against fields the client can already filter on and already read, so a forged cursor grants nothing forging a filter does not. **That rests on an invariant `QueryNormalizer` now enforces** — a cursor sort key must be on `sortable` ∩ `filterable` ∩ `selectable`. The first revision of this PR asserted "sortable, *therefore* filterable", which was false: the three allowlists are independent selectors, and gating on `sortable` alone made the cursor path a way around the other two in both directions. See the review-fix commits.

The base64 codec is hand-rolled and lives in one module (`packages/core/src/query/cursor.ts`) with its own tests. `packages/core/tsconfig.json` sets `"types": []` on purpose, so `btoa`/`TextEncoder` are as unavailable as `Buffer`. The payload is escaped to ASCII before encoding, so no UTF-8 encoder is needed — tested against accented Latin, CJK, and astral emoji.

**2. Union, not a shape change.** As above. `Pagination` is now generic (`<Entity = unknown>`) so `keyset` can be a typed `FilterExpression`; the default parameter keeps every bare `Pagination` reference compiling.

**3. Envelope `offset` on a cursor page → `0`,** since the field is non-nullable and normative. Documented in `ListResultDto`, in ADR-0021 §6, and in the adopter docs as "a keyset page knows what comes after a row, not how many precede it".

**4. Tiebreaker validation is in `QueryNormalizer`,** after sort resolution, on **both** `normalizeWire` and `normalizeInput` — `PaginationStrategy.normalize`'s signature is untouched. `QueryContext.cursor` is the programmatic input, and passing it under a non-cursor strategy is a `KAVO_QUERY_UNSUPPORTED_PARAM` rather than a silent no-op.

Rules enforced (`KAVO_QUERY_CONFLICTING_PARAMS` on `sort`): the effective sort must **end in `idField`** — took the `idField` reading as directed, since `EntityMetadata` carries no other uniqueness info, and documented the limitation rather than adding a config key — and every sort key must be a root scalar column (no relation paths, no `json`).

One deviation worth flagging, **and it is a known limitation rather than a safe fallback**: I first rejected **nullable** sort fields too, and had to back it out, because Mongoose reports every non-`required` path as nullable. The decoder's rule — a cursor may not carry `null` for any key — is only *half* a guard. It gives the documented loud 400 when NULLs sort first, but when they sort last (PostgreSQL `ASC`, sqlite `DESC`) the null-keyed rows are **silently omitted from every page** with no error, while `total` still counts them. ADR-0021 §4 now states both failure modes in a table and says plainly that cursor pagination must not be used over a nullable sort key; two fuller fixes are named and deferred. Both halves are pinned by tests against a real database.

**5a. `nextCursor` is computed from the entity, before serialization** — in the engine's `listMeta`, which receives the `FindManyResult`. There's a test proving `fields=name` (which strips `id` from the wire shape) doesn't break paging.

**5b. The `limit + 1` over-fetch lives in the built-in `findMany` handler**, not in the four adapters. The handler asks the adapter for one row past the page, drops the sentinel, and reports `FindManyResult.hasMore`; the engine turns that into `meta.nextCursor`. That keeps `EntityReader`'s contract as "return exactly what the query asks for" — a third-party adapter needs no cursor awareness beyond honoring `readFilter` — and avoids four copies of the same off-by-one. Cost is one extra row per page, never an extra query.

**6.** Uses #122's seam: `nextCursor` folds in through the private `listMeta()` step, not a top-level envelope field. The strategy's key is the *base* and a handler's `meta` merges over it, so a `withListMeta` contributor naming `nextCursor` still wins.

**7. Per-adapter keyset translation composes**, because there is none: core builds the predicate as an **ordinary filter AST** (`keysetExpression` — an `OR` of `AND` chains, flipping to `LT` per `desc` key), and each adapter's `findMany` calls `readFilter(query)` in place of `query.filter`. Every adapter already translates that AST, so composition with the client filter, include joins, and the soft-delete scope is free, and mixed `asc`/`desc` works everywhere at once. `count` deliberately keeps using `query.filter`, so `pagination.count`/`total` behavior is unaffected — asserted per adapter.

**8.** ADR-0021 (renumbered from 0019 — #127 keeps 0019 and #128 takes 0020) written, referenced from `05-query-grammar.md` §3, the configuration table, the adopter docs, the four adapter docs, `03-core-contracts-and-type-system.md`, `07-crud-engine.md`, and the code that implements it.

**9.** `offset` is still the default. Backward (`before`) paging and Relay conventions remain out of scope.

## Testing

**144 new tests** across core, the four adapters, and the two protocol bindings; `pnpm test` is 79 files / 1581 tests green.

`packages/core/tests/cursor-pagination.spec.ts` (77): the codec (round trips, non-ASCII, the URL-safe alphabet, and every rejection path — out-of-alphabet characters, the `%4 === 1` truncation guard, a valid-alphabet token whose bytes are not JSON, non-array payload, arity mismatch, type mismatch, enum outside the set, null key, unparseable date, an unknown sort field); `cursorValuesOf`'s encode-time type validation over `bigint`, a stringified number, and a `Decimal`-like object; `keysetExpression` at one, two and three keys including the new leading range bound; `readFilter`'s composition cases; the strategy; the normalizer's rules on both entry points — including four tests for the `sortable` ∩ `filterable` ∩ `selectable` invariant that **fail without the `resolveKeyset` fix** — and the engine end to end, covering the exact-exhaustion boundary, `limit === maxLimit`, a replacement handler omitting `hasMore`, a cursor emitted carrying `null` and 400-ing on replay, and the actual selling point: no shift when a row is inserted before the cursor, and clean resumption when the row the cursor names is deleted.

`cursor-pagination.spec.ts` per adapter (17 / 14 / 15 / 14 — TypeORM / Prisma / Mongoose / MikroORM): each walks a real database. Full traversal with no gaps or repeats, mixed asc/desc, **a date-typed cursor key with tied timestamps**, **a text key compared against a single unpaged read so the database's `ORDER BY` collation and the keyset's `>` are proven to agree**, a three-key sort whose deepest AND chain decides a boundary, composition with the client filter (plus `total` unaffected), with includes/populate, with soft delete, `withDeleted`, **and all of filter + include + soft delete combined in one walk**, plus tampered-cursor and missing-tiebreaker rejections asserting code and status. TypeORM additionally pins both halves of the null-sort-key limitation, using sqlite's opposite NULL orderings per direction.

`list-meta.spec.ts` gains `withListMeta` under cursor pagination, including the documented precedence where a contributor naming `nextCursor` wins. The GraphQL and MCP suites each gain three tests for the bootstrap refusal.

The in-memory core fixtures evaluate the comparison subset of the filter AST and honor sort, so the core end-to-end tests page for real rather than against a stub. The offset-only fixtures across `nest`, `graphql`, and `mcp` narrow the union — the consumer-side break, demonstrated.

**Verified gate** (run in this worktree):

```
> kavo@ depcruise
✔ no dependency violations found (310 modules, 1141 dependencies cruised)

> kavo@ test
 Test Files  79 passed (79)
      Tests  1581 passed (1581)
```

`pnpm check` green (generate + build + typecheck + depcruise + lint + test). `pnpm docs:links` → "doc links OK". `pnpm format:check` → "All matched files use Prettier code style!". `pnpm docs:build` → "build complete".


mahdavipanah:
## Review fixes pushed — 6 commits (`442b21c..8f18cb6`)

Every blocking finding from the reviewer, test-auditor, security-auditor and perf-auditor pass is addressed. Rebased onto `442b21c`; base unchanged (`feat/122-list-result-meta`).

### Security

**The cursor path was gated on `sortable` alone.** `resolveKeyset` never called `requireAllowlisted`, so `keysetExpression`'s `EQ`/`GT`/`LT` nodes reached the adapter through `readFilter` *after* `DefaultFilterParser` and `validateExpression` had run. With `filterable: ["status","id"]` and `sortable` at its default, `?filter[email][gt]=` was a 400 while `?sort=email,id&cursor=…` emitted `WHERE email > 'm' OR …` — a value-extraction oracle by binary search. Separately, `cursorValuesOf` reads the raw entity and `meta` never passes through the serializer, so excluding `passwordHash` via `selectable` or an item DTO while leaving `sortable` alone base64-encoded the hash into every `meta.nextCursor`. Both now run the same `requireAllowlisted` gate the filter and fieldset parsers use: the eligible set is **`sortable` ∩ `filterable` ∩ `selectable`**, and a failing key is **rejected**, not dropped — dropping one would break the total order.

ADR line 74 asserted "sortable, **and therefore** filterable" and built the whole no-signature argument on it. That was false. §2 now states the invariant as *enforced*, with both attack shapes spelled out.

### Performance

**The predicate was not index-startable.** The OR-of-ANDs shape is unchanged (it is the only shape that handles mixed asc/desc), but it is now AND-ed with a redundant non-strict bound on the leading key — `(a <= va) AND ( … )` descending, `>=` ascending. Logically implied by every branch, so no result changes; it just gives PostgreSQL/MySQL a real range qual instead of an ordered scan with the OR as a filter (or a BitmapOr that loses the ordering). Five lines in core, no adapter changes.

The `O(limit)` claim in `using-the-api.md` is now conditioned on **a matching composite index in that exact column order and direction**, with the note that an unindexed MongoDB sort *errors* at the 32 MB limit rather than merely slowing down. `pagination.count: false` is recommended alongside `strategy: "cursor"` with the reason stated; **the default is unchanged** — that would be a wire-contract change.

### Correctness

- **Null sort keys silently drop rows.** The ADR, the `resolveKeyset` docblock and the adopter docs all claimed paging "works until it reaches a null sort key and then fails loudly". That only holds when NULLs sort **first**. When they sort last (PostgreSQL `ASC`, sqlite `DESC`), the null-keyed rows are never returned, `nextCursor` goes to `null` as if the walk finished, and `total` still counts them. All three sites now say so plainly, in a table of both failure modes, and state that cursor pagination must not be used over a nullable sort key. Both halves are pinned against a real database in the TypeORM suite, using sqlite's opposite NULL orderings per direction. The two fuller fixes (a per-adapter nullability-trust signal; a NULL-aware keyset) are named and deferred — neither is a documentation change, and neither belongs in the change that made the documentation honest.
- **`bigint`/`Decimal` keys broke paging outright.** `cursorValuesOf` now validates each value against its declared `FieldKind` with the same predicate `reviveValue` decodes with, raising `ConfigurationException` naming the column and both types — instead of a `JSON.stringify` 500 on page 1 (MikroORM v7, Prisma) or a permanent page-2 400 blaming the wrong thing (TypeORM). The supported runtime representations are recorded in the ADR.
- **GraphQL and MCP silently returned wrong rows.** Both hard-code `limit`/`offset` and `normalizeInput` ignores `offset` under a keyset, so `books(limit: 20, offset: 40)` returned rows 1–20 undetectably, with `nextCursor` unreachable. Upgraded from "document it" to **fail at bootstrap**: `crudFields` and `crudTools` raise `ConfigurationException` naming the entity, the key and the way out. Recorded in ADR §7 and both doc sites. (Read through a structural probe rather than adding `engine` to `BoundKavoService`'s `Pick`, which would drag the entity-typed `ResolvedEntityConfig` into an invariant position and break `discovery.ts`.)
- **`?cursor=` was ignored on the wire under a non-cursor strategy** while the programmatic path rejected it and the docstring promised rejection. Both paths now share one check, and it probes the strategy **structurally** rather than comparing its name to `"cursor"` — a third-party keyset strategy is no longer silently downgraded.
- **`decodeCursor` threw a `TypeError`** on an unknown sort field; it now reports a query issue. It and `keysetExpression` are also **off the barrel** now (ADR-0010) — normalizer internals whose contracts only hold post-validation.
- **A no-progress guard** in `listMeta`: a partially-migrated adapter that kept `query.filter` instead of `readFilter(query)` looped a client forever. Erroring beats looping.

### Tests

The spec file's **literal NUL byte** is gone, so the PR diff is readable text (`926 ++++` rather than `Bin 0 -> 22267`). The fixture that hid both allowlist bugs now takes per-test overrides — **the four new allowlist tests fail without the `resolveKeyset` fix** (verified by reverting it: 4 failed / 72 passed).

144 new tests overall. Per adapter, all four: a date-typed cursor key with tied timestamps; a text key checked against a single unpaged read so collation and the keyset are proven to agree; a three-key sort whose deepest AND chain decides a boundary; **filter + include + soft delete combined in one walk**; and `withDeleted`. Core gains the untested codec branches, the encode-time validation cases, the null path end to end, insert/delete paging stability, `limit === maxLimit`, a handler omitting `hasMore`, and `withListMeta` + cursor including the documented precedence. Weak assertions (`toHaveLength(1)`, bare `toBeInstanceOf`, `detail`-substring-only) now assert codes and quoted field names, and the `adapter.lastQuery` over-fetch probe is replaced by the contract it stood in for.

### Docs

ADR renumbered **0019 → 0021** (#127 keeps 0019, #128 takes 0020) — file, heading, sidebar, and every citation. Five stale architecture docs updated: all four adapter docs still described `findMany` as plain skip/take with no `readFilter`; `07-crud-engine.md` omitted `hasMore` and described `meta.nextCursor` in the future tense; `03-core-contracts-and-type-system.md` still listed a scalar `Pagination`.

### Not changed, deliberately

- `pagination.count`'s default, `offset` as the default strategy, the OR-of-ANDs shape itself, backward paging / Relay conventions — all out of scope or explicitly excluded.
- Adding `cursor` + `meta` to the GraphQL/MCP surfaces — a schema change to two public protocols; failing fast is correct now and stays correct after.
- Renaming `readFilter` → `pagedReadFilter` — a public rename touching core plus four adapters plus docs, for a naming improvement, on an already-large diff. Worth a follow-up if wanted.

### Gate

```
✔ no dependency violations found (310 modules, 1141 dependencies cruised)
 Test Files  79 passed (79)
      Tests  1581 passed (1581)
```

`pnpm check` green. `pnpm docs:links` → "doc links OK". `pnpm format:check` → clean. `pnpm docs:build` → "build complete".

h16d (approved):

PR #205, opened 2026-08-13
feat(core)!: add search[query]/search[mode]/search[fields] query grammar

### LINKED ISSUE #156: Add search[query]/search[mode]/search[fields] query grammar for free-text search
## Context

Kavo's query grammar supports rich per-field filtering (`filter[field][operator]=value`) but has no ergonomic way to search a free-text term across multiple fields at once — the common "search box" use case. This adds a `search[...]` bracket-notation query param (parallel to `filter[...]`) that composes into the existing `Filter` AST as a synthesized `OR`-of-`ILIKE` group, `AND`-ed with any explicit filters.

Design is fully worked out in `docs/superpowers/specs/2026-08-08-search-feature-design.md`, extended by this issue with a `search.driver` setting (see below) that the spec's default did not name — this issue implements that spec plus the extension. No new `FilterOperator`, no adapter-level code changes (all four `FilterTranslator`s already handle OR-groups of LIKE/ILIKE/pattern conditions), no new ADR.

## Acceptance criteria

- [ ] `EntityConfig.searchable?: QueryFieldSelector<Entity>` added, resolved in `resolve-entity-config.ts` alongside `filterable`/`sortable`/`selectable`; defaults to all own scalar string columns when unconfigured (like `filterable`'s default); relation paths (e.g. `brand.name`) are accepted.
- [ ] `QuerySettings` gains `search.enabled` (default `false`) and `search.mode: 'substring' | 'words'` (default `'substring'`), resolved through the global → entity → operation → per-call precedence chain, added to `BUILT_IN_DEFAULTS`, `KavoSettings`, `validateSettings`, and `describeResolvedConfig`.
- [ ] `QuerySettings` also gains `search.driver: 'orm'` (default, and for now the only accepted value), resolved through the same precedence chain and added to the same `BUILT_IN_DEFAULTS`/`KavoSettings`/`validateSettings`/`describeResolvedConfig` set. This is a reserved discriminator, not a pluggable backend seam: it exists so a future `'postgres'` (native full-text) or `'meilisearch'` driver can land additively later without a breaking config change now. `validateSettings` rejects any value other than `'orm'` at bootstrap. `search.driver` is config-only — there is no `search[driver]` wire token; callers never choose the backend per-request.
- [ ] Wire grammar `search[query]=<term>&search[mode]=substring|words&search[fields]=<comma-list>` parsed in the filter parser / `QueryNormalizer`, synthesizing:
  - substring mode: one `OR` group of `ILIKE '%term%'` across searched fields
  - words mode: one `OR` group per whitespace-split word, `AND`-ed together
  - the synthesized fragment is `AND`-ed with any existing `filter[...]` conditions
- [ ] `search[fields]` narrows to a subset of the resolved `searchable` allowlist; unset means all of `searchable`.
- [ ] Validation: 400 when `search[query]` is supplied but `search.enabled` is `false`; 400 when `searchable` resolves empty (explicit `searchable: []`); 400 when `search[fields]` names a field outside `searchable`; 400 on an invalid `search[mode]` value; 400 when `search[mode]`/`search[fields]` is present without `search[query]`.
- [ ] Literal `%`/`_` in a search term are escaped, matched literally (reuses existing LIKE escape-and-bind logic).
- [ ] Unit tests in `packages/core/tests/` covering config resolution (including `search.driver` defaulting to `'orm'` and `validateSettings` rejecting any other value), parser/normalizer synthesis (both modes), field narrowing, and every validation case above.
- [ ] One test per ORM adapter (`packages/orms/{typeorm,prisma,mongoose,mikroorm}/tests/`) confirming correct translated output for both modes, a relation-path field, and a `search[fields]`-narrowed request.
- [ ] End-to-end Nest test (`packages/frameworks/nest/tests/`) exercising `search[query]`/`search[mode]` through a generated route.
- [ ] `docs/internals/architecture/05-query-grammar.md` gets a new `## 4. Search` section; `docs/internals/architecture/08-configuration.md` gets `searchable`/`search.enabled`/`search.mode`/`search.driver` added to the config key tables.
- [ ] `pnpm check` passes.

## Affected packages

`@kavo/core`, `@kavo/typeorm`, `@kavo/prisma`, `@kavo/mongoose`, `@kavo/mikroorm`, `@kavo/nest`, docs

## Constraints

- Reuses the existing `QueryFieldSelector` allowlist mechanism (`filterable`/`sortable`/`selectable` precedent) — no second config mechanism, per the `add-config-key` skill's normative rule.
- No new `FilterOperator` or AST node — `search[...]` compiles to existing `Filter`/`FilterGroup`/`ILIKE` nodes, so the operator mapping table in doc 05 §1 and each adapter's exhaustive `FilterTranslator` switch are unchanged.
- Wire tokens stay exact-case matched, no aliases (per doc 05's existing grammar rules).
- Config keys stay camelCase with positively-phrased booleans (`search.enabled`, not `search.disabled`) per `CLAUDE.md` conventions.

## Out of scope

- Full-text search (DB-native `tsvector`/GIN, ranking) and any other backend (e.g. Meilisearch) — deferred as a possible future `search.driver` value (`'postgres'`, `'meilisearch'`, …) once one is actually needed. Not `search.mode`: `mode` stays substring-vs-words semantics over the ORM path; a different backend is a different `driver`.
- Implementing any `search.driver` value other than `'orm'`, and any driver-specific behavior (native full-text query building, external search index sync, ranking). `search.driver` this issue adds is a reserved discriminator so those can land additively later, not a pluggable backend registry — building the registry itself is still out of scope and would warrant its own ADR if/when a second driver is implemented.
- GraphQL/MCP surface for `search[...]` (protocol bindings can pick it up in a follow-up once the core/REST grammar lands).

h16d:
## What / why

Adds a `search[query]=<term>` free-text search parameter (parallel to
`filter[...]`) that composes into the existing filter grammar rather than
introducing a new mechanism: it synthesizes a `Filter` AST fragment — an
`OR` group of `ILIKE '%term%'` conditions across an allowlisted field set
(`allowlists.searchable`), one `OR` group per word in `words` mode
(`AND`-ed together) — and `AND`s it into whatever `filter[...]` conditions
are already present. No new `FilterOperator`, no adapter-level changes:
every `FilterTranslator` already handled an `OR` group of `ILIKE`/pattern
conditions, verified empirically against all four adapters rather than
assumed.

`allowlists.searchable` follows the same `QueryFieldSelector` shape as
`filterable`/`sortable`/`selectable`, defaulting to every own **string**-kind
column (narrower than `filterable`'s "every own column", since a non-string
column has nothing an `ILIKE` fragment can usefully match); relation paths
are permitted. `query.search.enabled`/`.mode`/`.driver` resolve through the
standard global → entity → operation → per-call precedence chain,
`enabled` defaulting to `false` so `search[query]` 400s until an entity or
operation opts in. `search.driver` is a reserved discriminator (`'orm'` is
the only accepted value today) for a future pluggable search backend.

`search[mode]=words` is capped at `query.maxInValues` (reusing the existing
limit rather than adding a new config key) — word count × searched-field
count would otherwise synthesize an unbounded number of `ILIKE` conditions
from a single query string.

Three deliberate deviations from the linked design spec (documented in a
header on that doc, brought onto this branch since it wasn't on `main`):
`allowlists.searchable` (not a top-level `EntityConfig` key, following the
precedent `479d762` set for `includable`), `query.search.*` nesting (not a
top-level `search` settings key, per issue #156's own acceptance criteria),
and the string-kind-only default.

Closes #156

## Public API impact

Additive to `@kavo/core`'s barrel: `SearchMode`, `SearchDriver`,
`SearchSettings` types exported; `QuerySettings.search`,
`QueryAllowlists.searchable`, `ResolvedQueryAllowlists.searchable` added.
Marked `!` because `ResolvedQueryAllowlists` (and `KavoSettings.query`)
gain a required member — anyone constructing one of those shapes directly
(rather than through `resolveEntityConfig`) needs to add it — same
treatment `479d762` gave the equivalent `includable` addition. No changes
to any existing exported member's shape or behavior.

## Testing

- `packages/core/tests/`: config resolution (default string-column
  allowlist, explicit array/`{exclude}`, relation paths, computed-field
  rejection), `validateSettings` for `search.enabled`/`.mode`/`.driver`,
  `QueryNormalizer` synthesis (both modes, `search[fields]` narrowing,
  every validation case from the issue, escaping, AND-composition with
  `filter[...]`, the words-mode word-count cap, and that an unrelated
  filter issue doesn't raise a spurious search issue).
- `packages/orms/{typeorm,prisma,mongoose,mikroorm}/tests/`: one test file
  per adapter confirming the synthesized AST translates correctly for both
  modes, a relation-path field, and a narrowed single-field request.
- `packages/frameworks/nest/tests/`: end-to-end coverage through a
  generated route (default allowlist, both modes, the disabled-search 400).
- `pnpm build` / `typecheck` / `depcruise` / `lint` / `docs:links`: all
  green. `pnpm test`: 2143/2143 non-container tests pass. 5 testcontainers
  example e2e suites (Postgres/MariaDB/Cockroach/Mongo) fail in this
  sandbox because the Docker daemon isn't reachable — confirmed
  environmental (fails in `beforeAll`, before any app code runs, even with
  the sandbox disabled) and unrelated to this change; CI has Docker.

## Review notes

- Issue #156 carries both `status:ready` and `status:maybe` ("Might not be
  implemented — not yet committed") labels — flagging in case that wasn't
  intentional.
- Worth double-checking whether bundling the design-spec doc onto this
  branch (`docs/superpowers/specs/2026-08-08-search-feature-design.md`,
  not required by any link check or acceptance box) is wanted, or should be
  dropped.
- `search[...]` composing into `filter` on `findOne` (single-row-by-id
  reads) is parsed/validated but has no effect there, same as `filter[...]`
  already behaves today — the adapter's `findOneById` never applies
  `query.filter`. Consistent with existing behavior, not new.


h16d:
Ran the full reviewer fan-out (correctness/pnpm check, test-coverage, security, perf) and pushed fixes:

**Blocking, now fixed:**
- An explicit `allowlists.searchable` entry naming a non-string own column (e.g. `searchable: ['age']`) was accepted at bootstrap and only failed at request time as an unmapped driver error. Added a bootstrap `ConfigurationException` for own-column entries (relation-path entries stay unchecked — no target metadata in scope, matching `filterable`'s existing laxity there).
- `search[mode]=words` capped only the word count against `query.maxInValues`, leaving the field-count factor (which defaults to *every* own string column) unbounded — an entity with a wide `searchable` allowlist could still synthesize thousands of `ILIKE` conditions from a legal word count. Now caps the product (`terms × fields`).

**Also addressed** (non-blocking, but folded in): adopter-facing docs (`using-the-api.md`, the Nest configuration reference) now cover `search[...]` and `allowlists.searchable` — previously only the internal architecture docs did; added a note that `searchable` columns need a trigram/GIN index since every synthesized pattern has a leading wildcard; added tests for the operation-scope precedence chain on `query.search`, a repeated-key/array `search[query]` rejection, relation-path search synthesis through the normalizer, and `searchable`/`filterable` allowlist independence.

**Declined:** renaming `search.driver` to align with the existing `strategy` naming pattern — `driver` is the literal name issue #156's acceptance criteria specified.

`pnpm check`: build/typecheck/depcruise/lint green, 2152/2152 non-container tests pass. The 5 testcontainers example suites (Postgres/MariaDB/Cockroach/Mongo) still fail in this environment only because the Docker daemon isn't reachable — pre-existing on `main`, unrelated to this PR.

Security audit came back clean (no allowlist bypass, no injection, escaping verified, no disclosure-rule violation). Full findings available on request.

h16d:
Added two more pieces you asked about:

**Swagger docs for search[...]** — `search[query]`/`search[mode]`/`search[fields]` are now documented, listing the entity's actual searchable allowlist (\"Allowed fields: title.\"), the same as `filter`/`sort`/`fields` already were. This had to be deferred to `KavoModule`'s discovery binder rather than applied at `@Kavo` decoration time, for the same reason `applyConditionalRequestDocs` (the ETag/conditional-request docs) already is: `query.search.enabled` needs the full global→entity→operation precedence chain, which doesn't exist yet at decoration time. A nice side effect: because it runs after ORM metadata exists, the deferred `allowlists.searchable` description is the fully *resolved* array — better than `filter`/`sort`/`fields` get, which stay undescribed on an `{ exclude }` or default config. Params are omitted entirely when search isn't enabled, rather than advertised as something that would always 400.

**nest-typeorm example** — `search[query]` is now enabled on `Cat` (explicit `searchable: ["name"]`, alongside its existing explicit filterable/sortable/selectable lists) and `Owner` (left unconfigured, demonstrating the zero-config default — every own string column: `name` and `email`). Both get real-database e2e coverage: substring matching, composing with an existing `filter[...]`, and confirming search stays off on `Tag`, which never opted in.

`pnpm check`: build/typecheck/depcruise/lint green, `docs:links` green, 2157/2157 non-container tests pass (same 5 pre-existing Docker-unreachable example suites as before, unrelated).

Issue #291, opened 2026-08-26
CreateOne ignores ManyToOne relation UUID string — 500 instead of 201


h16d:
## Problem

With a `@ManyToOne` relation and `creatable: ["product"]`, sending `{ product: "<uuid>" }` in the create body produces a **500** (`KAVO_PERSISTENCE_FAILED` — NOT NULL constraint on the FK column) instead of resolving the relation and persisting.

Issue #150 documents this works on 0.7.2. On **0.14.5** it does not.

## Entity

```ts
@Entity("invoice")
export class Invoice {
  @PrimaryGeneratedColumn("uuid") id!: string;

  @ManyToOne(() => User, (u) => u.invoices, { nullable: false })
  @JoinColumn({ name: "user_id" })
  @CreatedBy()
  user!: User;

  @ManyToOne(() => Product, { nullable: false })
  @JoinColumn({ name: "product_id" })
  product!: Product;
}
```

## Controller

```ts
@Kavo(Invoice, {
  operations: { createOne: true },
  allowlists: {
    creatable: ["product"],
    // ...
  },
})
```

## Request / Response

```
POST /invoices
Authorization: Bearer <token>
Content-Type: application/json

{ "product": "00000000-0000-0000-0000-000000000001" }

→ 500 KAVO_PERSISTENCE_FAILED
  "NOT NULL constraint failed: invoice.product_id"
```

Expected: 201 with the invoice created and `product_id` set to the supplied UUID.

## Notes

- `creatable: ["product_id"]` with a scalar `product_id` column works fine — the issue is specific to the relation path.
- Sending `{ product: { id: "<uuid>" } }` also fails.
- ADR-0014 says the relation name should be the write spelling. It appears the TypeORM adapter does not yet resolve a UUID string to the `@JoinColumn` during createOne.

h16d:
Verified fixed: `bcf9263` ("fix(core): reject bare scalar id for single-key relation association (#293)") is already on `main`. A bare scalar sent for a single-key relation now raises `AssociationInvalidShapeException` (400) instead of reaching the adapter and failing the FK NOT NULL constraint as a 500. `packages/core/tests/serializer.spec.ts` covers it (46 passing). Closing.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #293, opened 2026-08-26
fix(core): reject bare scalar id for single-key relation association (issue #291)

h16d:
ADR-0014 previously accepted `{"product": "<uuid>"}` as shorthand for
`{"product": {"id": "<uuid>"}}` on create/update/patch. The shorthand
left intent ambiguous and, resolved wrong, surfaced as an opaque
FK-constraint failure instead of a clear 400. `DefaultDeserializer`
now requires the reference-object form for a single-key relation and
throws AssociationInvalidShapeException (KAVO_ASSOCIATION_INVALID_SHAPE,
400) for a bare scalar. Composite-key targets (ADR-0039) are unaffected:
they have no single column a scalar could be mistaken for, so their own
`~`-delimited shorthand still works.

Updates every write path that shares associate() (replace/resource/
jsonPatch array-mutation strategies) and the example apps' DTOs/tests
to the reference-object shape.


h16d:
`build (node current)` and `test (node current)` are failing on `better-sqlite3`'s native `node-gyp rebuild` step:

```
gyp http 404 https://nodejs.org/download/release/v26.8.0-alpha.0.0.0/node-v26.8.0-alpha.0.0.0-headers.tar.gz
```

`actions/setup-node@v7`'s `node-version: current` resolved to a nightly/alpha build (`26.8.0-alpha.0.0.0`) whose prebuilt headers aren't published on nodejs.org yet — a native-module compile failure with no relation to this PR's diff (core's association-shape change, mongoose test/DTO updates). It reproduces identically on `main`'s latest CI run (`8a37643`, run #552, also `failure` on both `node current` jobs), so it's a pre-existing, base-branch-red issue in the Node "current" release channel, not something introduced here. No fix exists yet within this PR's scope — pinning or working around the "current" matrix entry is a `.github/workflows/ci.yml` change unrelated to issue #291, so I'm not widening this PR to include it. Re-running both jobs once to confirm it isn't a one-off.

Every other check (build/test on `lts/-1` and `lts/*`, `test (bun)`, `coverage`, `format`, `docs`, `doc links`) is green on this PR's head.

---

PR #383, opened 2026-09-04
refactor(core)!: give cache one spelling for "off" — ttl presence is the switch

### LINKED ISSUE #377: Give `cache` one spelling for "off": make `ttl` absence the switch, reject `ttl: 0`
### Context

`cache` is `CacheSettings | false` and `ttl` doubles as the on/off switch: `ttl: 0` (the default) means the result cache is off, any positive `ttl` means on — there is no `enabled` key. That gives "off" two different spellings with two different meanings: `cache: false` disables the whole subtree (result cache **and** ETags), while `cache: { ttl: 0 }` disables only the result cache and silently leaves `etag: true`. `0` is a magic value that looks like a boundary number, and in the merge chain `operations: { findMany: { cache: { ttl: 0 } } }` reads like a disable but is really just assigning zero — a later per-call `cache: { ttl: 30 }` merges straight back over it. This makes `ttl` optional so **absence** is the switch and `0` is a validation error, leaving exactly one spelling per state.

### Proposed shape

```ts
export interface CacheSettings {
  readonly ttl?: number;   // absent → result cache off; must be a positive integer when present
  readonly etag: EtagSettings;
}
```

| Intent | Today | After |
|---|---|---|
| Everything off (no cache, no ETags) | `cache: false` | `cache: false` |
| Result cache off, ETags on | `cache: { ttl: 0 }` *or* `cache: false`+re-add | `cache: { etag: true }` (omit `ttl`) |
| Result cache on, 60s | `cache: { ttl: 60 }` | `cache: { ttl: 60 }` |

`BUILT_IN_DEFAULTS.cache` becomes `{ etag: true }` (no `ttl`). It stays a full object rather than `false` for the same reason it is one today — a partial `cache: { ttl: 60 }` override merges against a complete base instead of replacing `false` wholesale (`merge-settings.ts` comment, `defaults.ts`).

### Open design question — overriding an inherited `ttl` downward

With `ttl: 0` gone, how does a narrower scope turn a result cache **off** that a broader scope turned on, *without* also killing ETags (which `cache: false` does)? `mergeSettings` does not treat `{ ttl: undefined }` as a delete. Options for the ADR to pick:

1. Accept the gap — `cache: false` then `cache: { etag: true }` at the same scope is the answer; the "on globally, off for one operation, ETags kept" case is niche.
2. Allow `cache: { ttl: false }` (or `null`) as an explicit "override off" token — one sentinel, but a typed one that cannot be confused with a duration.
3. Keep `mergeSettings` treating an explicit `ttl: undefined` in an override as "clear it".

Recommendation: option 2 — `ttl: number | false`, `false` meaning "off, and override any inherited duration"; still one number-or-off shape, no magic `0`.

### Acceptance criteria

- [ ] `CacheSettings.ttl` is optional (or `number | false` per the design decision); `BUILT_IN_DEFAULTS.cache` is `{ etag: true }`.
- [ ] `validate-settings.ts` rejects `ttl: 0` and any non-positive / non-integer `ttl` with a message pointing at `cache: false` (whole subtree) vs omitting `ttl` (result cache only).
- [ ] Engine gate `cache === false || cache.ttl <= 0 ? null : cache` (`kavo-engine.ts:585`) and the "when `settings.cache.ttl` is positive" read (`resolved-entity-config.ts:93`) updated for the new shape.
- [ ] Whatever the design question resolves to is implemented and has a merge-chain test (broad scope on, narrow scope off, ETag state asserted both ways).
- [ ] `pnpm check` passes; tests covering `ttl: 0` as "off" are rewritten to the new spelling; a test pins that `ttl: 0` is now a bootstrap `ConfigurationException`.
- [ ] `feat!` commit with a `BREAKING CHANGE:` footer: `cache: { ttl: 0 }` no longer validates — use `cache: false` or omit `ttl`.
- [ ] ADR-0031 amended (or a new ADR) recording that `ttl` presence is the switch and why `0` was retired; doc comments in `settings.ts`, `defaults.ts`, `merge-settings.ts` updated.
- [ ] Docs: `docs/reference/config-keys.md`, `docs/guides/configuration/{settings,etag-overrides}.md`, `docs/features/{caching-and-etags,result-cache}.md`.

### Affected packages

`@kavo/core`, docs (and any `examples/*` app setting `cache.ttl`).

### Constraints

- Core stays dependency-free and owns its contracts (ADR-0001/ADR-0005).
- One settings precedence chain, `DeepPartial`-partializable, merge-compatible (`KavoSettings` schema-extensibility rule).
- `etag` still defaults `true` and still serves independently of the result cache (ADRs 0020/0027/0031).
- `false` still disables the whole `cache` subtree — unchanged.
- Config keys camelCase, booleans phrased positively (`CLAUDE.md` Conventions).

### Coordination

Touches `settings.ts`, `BUILT_IN_DEFAULTS`, `validate-settings.ts`, `config-keys.md` — same as #374 / #375 / #376. Sequence to avoid conflicts.

### Out of scope

- The cache store registration (`KavoOptions.cacheStore`) and invalidation model (ADR-0031) — unchanged.
- `etag`'s own shape.
- Per-key TTL / staleness analysis.

h16d:
## What & why

`cache: false` and `cache: { ttl: 0 }` both meant "off" but disabled different things — `false` disabled the whole subtree (result cache **and** ETags), while `ttl: 0` disabled only the result cache and silently left `etag: true`. `CacheSettings.ttl` is now optional (`number | false`, no longer defaults to `0`): an omitted `ttl` is the result cache's "off" spelling, `ttl: 0` fails bootstrap validation with a `ConfigurationException`, and `ttl: false` is the one explicit sentinel for overriding an *inherited* `ttl` back off at a narrower scope without disabling that scope's `etag` (since `mergeSettings` replaces keys an override supplies rather than clearing them). `BUILT_IN_DEFAULTS.cache` is now `{ etag: true }`. ADR-0031 is amended accordingly, and the config-keys reference, settings guide, result-cache feature doc, and configuration architecture doc are synced.

Closes #377

## Public API impact

`packages/core/src/index.ts` barrel is unchanged (no new/removed exports), but `CacheSettings.ttl`'s type changes from `number` to `number | false | undefined` — a breaking change for any consumer setting `cache: { ttl: 0 }`, which now fails bootstrap validation instead of meaning "off".

## Testing

Added: a bootstrap-rejection test for `cache: { ttl: 0 }`, and a merge-chain test proving `cache: { ttl: false }` overrides an inherited `ttl` back off while leaving `etag` on. Updated the existing `config-validation.spec.ts` malformed-`ttl` cases to include `0` and accept `false`.

`pnpm check`: build, typecheck, depcruise, and lint all pass clean. Test suite: 2942 passed, 633 skipped; the only 12 failing suites are pre-existing sandbox-environment failures unrelated to this change (no container runtime for testcontainers-backed Postgres/MariaDB/CockroachDB/Mongo e2e suites, and the MongoDB memory-server binary download is blocked in this environment). `pnpm docs:links` passes.

## Review notes

The design resolves the issue's "open design question" using its own recommended option (`ttl: number | false`, `false` as the override-off sentinel) rather than the two alternatives it also listed — worth double-checking that's the intended resolution before merge.

---
_Generated by [Claude Code](https://claude.ai/code/session_016z5oRHYR3ug1o5TXGKjW2e)_

## Summary by CodeRabbit

- **New Features**
  - Cache configuration now supports `ttl: false` to disable result caching while preserving ETag generation.
  - Omitted TTL values leave result caching disabled by default.
  - Positive integer TTL values enable result caching.

- **Bug Fixes**
  - Invalid TTL values, including `0`, non-positive numbers, and non-integers, are now rejected with configuration errors.
  - Cache overrides now correctly distinguish disabling result caching from disabling the entire cache configuration.

- **Documentation**
  - Updated cache configuration, reference, and architecture guidance with the revised TTL behavior and examples.


h16d:
Re the "Linked Issues check" warning:

- **`docs/features/caching-and-etags.md` / `docs/guides/configuration/etag-overrides.md`**: checked both — neither mentions `ttl` at all, only `cache.etag`, which this change doesn't touch. There's nothing in them to update; the issue's doc checklist listed every file under `docs/features`/`docs/guides/configuration` that talks about `cache`, but these two are ETag-only.
- **Commit convention**: the change does carry a `BREAKING CHANGE: cache.ttl is now optional and no longer defaults to 0. cache: { ttl: 0 } no longer validates; ...` footer (see `7189b44`), which is what release-please keys off. It intentionally does **not** use the `feat!`/`refactor!` header marker — this repo's own `wagoid/commitlint-github-action` job bundles a different (older) `@commitlint/*` than the one `pnpm install` gives you, and that bundled parser fails to extract `type`/`subject` from any `type(scope)!: subject` header, even though the repo's own pinned CLI (and `tests/commitlint-config.spec.ts`) accepts it fine. Confirmed by reproducing the failure locally and re-running `pnpm exec commitlint` against the amended message. Also, per this repo's `type:` vocabulary, the issue's `type:refactor` label maps to `refactor`, not `feat`.
- **"incorrect defaults and inheritance behavior" in the risk note**: I re-read `settings.md`/`config-keys.md` against the actual implementation and the merge-chain test (`cache.spec.ts`, "'cache: { ttl: false }' overrides an inherited ttl back off without disabling etag") — the docs match. If there's a specific line you think is wrong, please point at it and I'll fix it.

---

PR #398, opened 2026-09-05
feat: accept the { exclude } form in create/update fields config (#397)

### LINKED ISSUE #397: Accept the `{ exclude }` form in `create`/`update` config, matching the read-side field groups
## Context

\`EntityConfig.create.fields\` and \`EntityConfig.update.fields\` are positive-only allowlists (\`readonly FieldPath<Entity, 1>[]\`). Every read-side field group — \`filter.fields\`, \`sort.fields\`, \`select.fields\`, \`include.fields\` — also accepts an \`{ exclude: [...] }\` form meaning "every eligible field except these". A caller who wants to open all writable fields but one currently has to enumerate the entire writable projection by hand and keep it in sync as the entity grows. This adds the same \`{ exclude }\` option to the two write slots.

## Acceptance criteria

- [ ] \`WriteFieldsConfig<Entity>['fields']\` accepts \`readonly FieldPath<Entity, 1>[] | { readonly exclude: readonly FieldPath<Entity, 1>[] }\`.
- [ ] \`{ exclude }\` is resolved at bootstrap against the entity's derived writable projection (every non-generated scalar column except the primary key, plus relations — ADR-0014), producing the same resolved writable-field list an explicit array would.
- [ ] An \`exclude\` entry that is not a writable field of the entity is a bootstrap \`ConfigurationException\`, consistent with how \`resolveFieldSelector\` rejects unknown names for the read groups.
- [ ] \`{ exclude: [] }\` resolves to the full writable projection (equivalent to omitting \`create\`/\`update.fields\`).
- [ ] Precedence is unchanged: a registered \`dto.create\` / \`dto.update\` class still replaces and wins over the shorthand (whichever form it takes); \`create.default\` / \`create.apply\` / \`update.default\` / \`update.apply\` still compose exactly as today.
- [ ] \`rejectComputedWriteDtoKeys\` and \`dtoClassFromFields\` handle the \`{ exclude }\` form (a computed field appearing via the resolved list is still rejected).
- [ ] \`@kavo/nest\` OpenAPI component schemas (\`UserCreate\` / \`UserUpdate\`) reflect the resolved field set — verified, not assumed, since the schema is built from the synthesized DTO shape.
- [ ] Unit tests in \`packages/core/tests/\` cover: exclude resolution, unknown-name rejection, empty-exclude, interaction with \`default\`/\`apply\`, and DTO-class precedence. Type-level tests in \`packages/core/tests/types/\` pin the accepted config shape.
- [ ] Docs updated: the \`create\`/\`update\` write-scoping guide under \`docs/guides/\` (and any \`docs/core/\` reference for \`EntityConfig\`) documents the \`{ exclude }\` form alongside the array form.

## Affected packages

\`@kavo/core\`, docs. (\`@kavo/nest\` only if the OpenAPI check reveals the schema is not derived from the resolved shape.)

## Constraints

- **ADR-0014** — the writable projection definition is the universe \`exclude\` resolves against; do not widen it.
- **ADR-0048** — \`apply\` composition for \`create\`/\`update\` is unchanged; \`exclude\` narrows the allowlist only, it is not a mandatory constraint.
- **ADR-0026** — \`select\`/response confidentiality is separate; excluding a field from writes does not hide it from responses.
- **CLAUDE.md conventions** — config keys camelCase; reuse the existing \`{ exclude }\` shape and the \`resolveFieldSelector\` resolution path rather than inventing a second one.
- Bootstrap-time resolution only (ADR-0012): \`exclude\` is resolved once when ORM metadata is available, never per request.
- Core stays dependency-free (ADR-0005).

## Out of scope

- Adding \`{ exclude }\` to \`patch\` / \`item\` / \`list\` \`dto\` shorthand (\`FieldsShorthand\`) — those are a separate slot type.
- Per-operation (\`operations.createOne.*\`) write-field scoping beyond what already exists.
- Any change to \`create.default\` / \`create.apply\` semantics.
- \`*Many\` bulk write operations (#137).

h16d:
## What & why

`EntityConfig.create.fields` and `update.fields` only accepted a positive allowlist array, so opening every writable field but one meant enumerating the whole writable projection by hand and keeping it in sync as the entity grows. This adds the `{ exclude: [...] }` form the read-side field groups (`filter.fields`/`sort.fields`/`select.fields`/`include.fields`) already have — "every writable field except these".

It resolves at bootstrap against the ADR-0014 writable projection, which is now derived once by a shared `derivedWritableFieldNames` helper and reused by `DefaultDeserializer` so the resolution and the runtime enforcement can't drift. Following `resolveIncludableSelector`'s strict precedent, an `exclude` entry that names nothing writable is a `ConfigurationException`; `{ exclude: [] }` — or any exclude that removes nothing — resolves to the entity-derived default, exactly as omitting the key would (returning an explicit full list would otherwise widen writes by the soft-delete marker column).

On the nest side, the bind-time fallback `@ApiBody` schema now reads its field lists from the engine's resolved DTO resolver instead of a decoration-time reconstruction, so a `{ exclude }` form's generated `UserCreate`/`UserUpdate` schema shows the narrowed set the engine enforces.

Closes #397

## Public API impact

None. `WriteFieldsConfig['fields']` widens from `readonly FieldPath[]` to `readonly FieldPath[] | { readonly exclude: readonly FieldPath[] }` — additive, no barrel changes, not breaking.

## Testing

- `packages/core/tests/config.spec.ts` — exclude resolution, relations in the universe, `{ exclude: [] }` equivalence, unknown-name rejection, `default`/`apply` composition, registered-DTO precedence.
- `packages/core/tests/apply.spec.ts` — `create.apply` still forces a value for an excluded field; `create.default` does not.
- `packages/core/tests/types/write-fields-exclude.test-d.ts` — accepted config shapes plus `@ts-expect-error` guards.
- `packages/frameworks/nest/tests/binding.e2e.spec.ts` — generated POST/PUT body reflects the narrowed set.
- `pnpm check` and `pnpm docs:links` both green (136 files, 3641 tests).

## Review notes

- Sharp edge, documented in `docs/features/allowed.md` rather than fixed (issue's out-of-scope list forbids changing `create.default`/`apply` semantics): `create.default`/`update.default` fills gaps *within* the field allowlist, so a default for a field `{ exclude }` removed never fires — `apply` is the tool for a server-set value on an excluded field. Pre-existing for the array form; `{ exclude }` makes it easy to reach.
- `packages/core/src/serialization/default-serializer.ts` is touched only to call the extracted `derivedWritableFieldNames` — behaviour identical, one derivation instead of two.

## Summary by CodeRabbit

* **New Features**
  * Added support for excluding specific writable fields from create and update configurations.
  * Exclusions are resolved at startup, with invalid field names producing configuration errors.
  * API documentation reflects resolved writable fields, including exclusion-based configurations.
  * Clarified interactions with defaults, applied values, relations, and registered DTOs.

* **Bug Fixes**
  * Ensured excluded fields are handled consistently across validation, defaults, and applied values.
  * Improved handling of ORM-derived fields: excluded by default and available only when explicitly selected.


h16d:
Superseded by #400, which carries all of this PR's commits plus the fixes needed to get CI green (a build-breaking naming mismatch and a derived-field write regression). Closing this in favor of #400.

---

Issue #373, opened 2026-09-04
Replace `computed` fields with ORM-metadata-derived virtual fields (filterable, sortable, selectable)


h16d:
### Context

Kavo's `computed` feature (ADR-0019) is a response-only escape hatch: a `resolve(entity, context)` function evaluated per row by `DefaultSerializer`, deliberately never filterable, sortable, or writable because it has no SQL behind it. Every supported ORM already has its own virtual/generated-column mechanism, and two of them (TypeORM `@VirtualColumn`, MikroORM `@Formula`) expose the derived SQL expression as schema metadata.

This issue removes `computed` from Kavo and instead lets the ORM own the definition. Kavo reads the derived expression from ORM metadata and — where the expression is real SQL — makes the field a first-class member of the `filterable` / `sortable` / `selectable` allowlists, translating predicates and sort keys by inlining the expression. Where the ORM's virtual is JS-only (Mongoose virtuals, Prisma client-extension fields), the field is response-only at best and filter/sort on it is a documented 4xx.

### Acceptance criteria

- [ ] `ComputedFieldDescriptor` / `ComputedFieldMap`, the `computed` config key, `resolveComputedFields`, `rejectComputedWriteDtoKeys`, and all `resolved.computed` wiring are removed from `@kavo/core`.
- [ ] `DefaultSerializer` / `DefaultDeserializer` no longer take or reference a computed map; the core barrel no longer exports the computed types.
- [ ] `FieldMetadata` (core metadata seam) gains an optional, ORM-opaque `derivedExpression` marker; core never inspects its contents (ADR-0005).
- [ ] `@kavo/typeorm` populates `derivedExpression` from `@VirtualColumn`'s `query` function; the adapter inlines it into `WHERE` / `ORDER BY` / `SELECT` when a filter, sort, or `select=` names the field.
- [ ] `@kavo/mikroorm` populates `derivedExpression` from `@Formula`; filter/sort/select on the field works (MikroORM supports `@Formula` in `where`/`orderBy` natively).
- [ ] `@kavo/prisma` and `@kavo/mongoose` expose no `derivedExpression`; a filter or sort targeting such a field returns a deterministic 4xx (`KAVO_QUERY_UNSUPPORTED` or equivalent), documented in the support matrix.
- [ ] A derived field is opt-in to `filterable` / `sortable` / `selectable` via `allowlists` — ORM metadata supplies shape only, never permission (same rule as relations).
[...]

PR #400, opened 2026-09-06
fix(core): exclude derived fields from the writable-field universe

### LINKED ISSUE #397: Accept the `{ exclude }` form in `create`/`update` config, matching the read-side field groups

h16d:
## What & why

This branch carries PR #398's `{ exclude }` write-field feature (`feat/397-create-update-fields-exclude`, merged in unchanged) plus one fix on top that gets it to a green gate.

PR #398 failed every CI job:
1. **Build failure** — `resolve-entity-config.ts` and `default-serializer.ts` imported `entityWritableFieldNames`, but the shared helper the PR extracted is actually named `derivedWritableFieldNames`. `tsc -b` failed outright.
2. **Real regression once the build compiled** — `derivedWritableFieldNames` only filtered out `field.generated` columns, not `field.derivedExpression` fields (the ORM-derived/virtual columns from issue #373, e.g. `@VirtualColumn`). That let a create/update body write directly to a computed field such as `fullName` or `slug`, violating the exact ADR-0014 writable-projection invariant the PR's own doc comment says the resolver and runtime enforcement must not drift on. Two tests in the PR's own suite caught this once the build was fixed:
   - `packages/core/tests/serializer.spec.ts` — deserializer accepted a `fullName` write
   - `packages/frameworks/nest/tests/binding.e2e.spec.ts` — a `slug` write reached the adapter

Fixed by excluding `derivedExpression !== undefined` fields in `derivedWritableFieldNames` alongside the existing `generated` check, and updated its doc comment.

A third, unrelated CI failure (`docs` job — 4 dead links to `/features/computed-fields`) predates this PR: it's leftover from #396, already merged to `main`, which deleted that page without updating its referrers. Fixed alongside since it was blocking a green PR.

Supersedes #398 (closed) — same feature diff, plus these fixes.

Closes #397

## Public API impact

None. Same as #398: `WriteFieldsConfig['fields']` widens from `readonly FieldPath[]` to `readonly FieldPath[] | { readonly exclude: readonly FieldPath[] }` — additive, no barrel changes, not breaking.

## Testing

- `pnpm check`: build, typecheck, depcruise, lint all green; 2970 tests passed, 0 failed (the 12 failing suites need Docker/testcontainers or a MongoDB binary download, both unavailable in this environment — unrelated to this change).
- `pnpm docs:links`: OK
- `pnpm docs:build`: OK

## Review notes

- The build-breaking rename and the derived-field leak are both isolated to `derivedWritableFieldNames` in `packages/core/src/metadata/entity-metadata.ts` and its two call sites — worth a close look there.
- Did not address CodeRabbit's "out of scope changes" flag on #398 about the computed-field-to-virtual-field refactor: that refactor was already merged into `main` via #396 before #398 branched, so it isn't actually part of this diff — CodeRabbit's comparison was against a stale base.

https://claude.ai/code/session_01V4WQTgFh6EiwzSSiXoUyvz


h16d:
The `docs` CI job failed on 4 dead links to `/features/computed-fields`, but this isn't this PR's regression — that page was deleted by #396 (already on `main`) when computed fields were replaced by ORM-derived virtual fields, and 3 other pages plus one reference doc still pointed at it / described the removed `computed` config key. `main` itself would fail `pnpm docs:build` the same way.

Fixed in c164e25: repointed `docs/core/entities.md`, `docs/guides/configuration/entity-config.md`, and `docs/reference/config-keys.md` at `docs/features/virtual-fields.md`, and dropped the stale `computed` key mention from `docs/reference/config-keys.md`. `pnpm docs:build` and `pnpm docs:links` are both green locally now.

---

